PR #5, opened 2015-05-29
composer.json is located in /opt/app/frontend

JamborJan:
```composer.json``` is located in ```/opt/app/frontend``` at least for Paperwork.
Added composer self update in case an old version needs to be updated on a local build machine.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -212,11 +212,13 @@ set -eu
 
 cd /opt/app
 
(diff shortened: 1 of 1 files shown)


JamborJan:
As discussed here: https://github.com/zarvox/vagrant-spk/issues/4

We should rather update the read-me and add examples to the config files.

Will close this for now and add some examples to https://github.com/zarvox/vagrant-spk/issues/4.

PR #7, opened 2015-06-06
Solves /opt/app.sandstorm/

JamborJan:
Solves https://github.com/zarvox/vagrant-spk/issues/6

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 `vagrant-spk` is a tool designed to help app developers package apps for [Sandstorm](https://sandstorm.io).
 
-Example usage:
+See also https://github.com/sandstorm-io/sandstorm/wiki/Porting-an-app-with-vagrant-spk for more details.
+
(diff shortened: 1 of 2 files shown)


paulproteus:
Thanks for the patch! I'll let @zarvox review it in detail, but it looks basically definitely correct.

I wonder how @rchrd2 was able to use `vagrant-spk` a few days ago!

paulproteus:
Ah hah, now I see -- the bug was introduced recently.

JamborJan:
```vagrant-spk dev``` was working fine but ```vagrant-spk pack``` not in this case.

Anyway, there seem to be some weird things ... still experimenting with https://github.com/JamborJan/paperwork/issues/10 and have no idea where it comes from.

zarvox:
Awesome, thanks a bunch!  Sorry for introducing a regression in one of my recent patchsets, and thanks for fixing it!

PR #11, opened 2015-06-11
Add "diy" stack, which has noops for each of the three scripts

### LINKED ISSUE #10: Create a "diy" vagrant-spk platform stack
Right now, if you want to package something where you want more control over how the VM gets brought up, etc,, you have to pretend you want `lemp` or something, and then remove things until you discover you are happy.

By contrast, we could give a minimal thing that doesn't work and tell you to customize it. That would be a `diy` platform stack.

(Are we OK consistently referring to these as "platform stack"s?)

zarvox:
@paulproteus Feel free to reword anything you like here.  When you're satisfied, click the merge button.

Closes #10.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -395,6 +395,86 @@ mkdir -p /var/run
 /usr/sbin/nginx -g "daemon off;"
 """
 
+DIY_SETUP_SCRIPT = r"""#!/bin/bash
+# This script is run in the VM once when you first run `vagrant-spk up`.  It is
+# useful for installing system-global dependencies.  It is run exactly once
+# over the lifetime of the VM.
+#
+# This is the ideal place to do things like:
+#
+#    export DEBIAN_FRONTEND=noninteractive
+#    apt-get install nginx nodejs nodejs-legacy python2.7 mysql-server
+#
+# If the packages you're installing here need some configuration adjustments,
+# this is also a good place to do that:
+#
+#    sed --in-place='' \
+#            --expression 's/^user www-data/#user www-data/' \
+#            --expression 's#^pid /run/nginx.pid#pid /var/run/nginx.pid#' \
+#            /etc/nginx/nginx.conf
+
+# By default, this script does nothing.  You'll have to modify it as
+# appropriate for your application.
+exit 0
+"""
+
+DIY_BUILD_SCRIPT = r"""#!/bin/bash
+# This script is run in the VM each time you run `vagrant-spk dev`.  This is
+# the ideal place to invoke anything which is normally part of your app's build
+# process - transforming the code in your repository into the collection of files
+# which can actually run the service in production
+#
+# Some examples:
+#
+#   * For a C/C++ application, calling
+#       ./configure && make && make install
+#   * For a Python application, creating a virtualenv and installing
+#     app-specific package dependencies:
+#       virtualenv /opt/app/env
+#       /opt/app/env/bin/pip install -r /opt/app/requirements.txt
+#   * Building static assets from .less or .sass, or bundle and minify JS
+#   * Collecting various build artifacts or assets into a deployment-ready
+#     directory structure
+
+# By default, this script does nothing.  You'll have to modify it as
+# appropriate for your application.
+cd /opt/app
+exit 0
+"""
+
+DIY_LAUNCHER_SCRIPT = r"""#!/bin/bash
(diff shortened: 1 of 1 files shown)


vagrant-spk
```diff
@@ -411,6 +491,11 @@ STACK_PLUGINS = {
         "setup": STATIC_NGINX_SETUP_SCRIPT,
         "launcher": STATIC_NGINX_LAUNCHER_SCRIPT,
     },
+    "diy": {
+        "setup": DIY_SETUP_SCRIPT,
+        "build": DIY_BUILD_SCRIPT,
+        "launcher": DIY_LAUNCHER_SCRIPT,
```
paulproteus: I'd like to see `launcher` get renamed to `start-grain` or similar, but that's not a show-stopper. I filed #15 about this wishlist desire of mine.

vagrant-spk
```diff
@@ -395,6 +395,86 @@ mkdir -p /var/run
 /usr/sbin/nginx -g "daemon off;"
 """
 
+DIY_SETUP_SCRIPT = r"""#!/bin/bash
+# This script is run in the VM once when you first run `vagrant-spk up`.  It is
```
paulproteus: I propose we add a one-line summary to the top of this, e.g.:

`Purpose: Install dependencies needed to run the app but that don't change when the source code changes`

But is that an accurate summary? In a way, the only difference between the `setup` and `build` phases is that `setup` is only run once, which seems to me to be an optimization.

I wonder if there's a clearer way to express this, and/or if you agree with the `Purpose: ...` line I named above.

paulproteus: Perhaps we could write a warning, like, 

Given that `setup` is run exactly once, if you change the contents of this file, you will have to call `vagrant destroy`, or execute the script manually within the VM. If that sounds like too much work, use `build` instead.

(as a proposal)

vagrant-spk
```diff
@@ -395,6 +395,86 @@ mkdir -p /var/run
...
+#            --expression 's#^pid /run/nginx.pid#pid /var/run/nginx.pid#' \
+#            /etc/nginx/nginx.conf
+
+# By default, this script does nothing.  You'll have to modify it as
+# appropriate for your application.
+exit 0
+"""
+
+DIY_BUILD_SCRIPT = r"""#!/bin/bash
+# This script is run in the VM each time you run `vagrant-spk dev`.  This is
```
paulproteus: Proposed addition to top:

`Purpose: Prepare the app code to be run, for example compiling it`

I am personally waffling about if I would put a `pip install`-type step in here, versus in the `setup`, but that mostly doesn't matter given that the Python platform support actually exists, and this is for DIY people anyway.

vagrant-spk
```diff
@@ -395,6 +395,86 @@ mkdir -p /var/run
...
+#     directory structure
+
+# By default, this script does nothing.  You'll have to modify it as
+# appropriate for your application.
+cd /opt/app
+exit 0
+"""
+
+DIY_LAUNCHER_SCRIPT = r"""#!/bin/bash
+# This script is run every time an instance of our app - aka grain - starts up.
```
paulproteus: `Purpose: Start the app on port 8000`

paulproteus:
This is good, and I will merge it before my non-showstopper comments turn into bikeshedding!

vagrant-spk
zarvox: You make an interesting point about the similarity between `setup.sh` and `build.sh` -- I suspect the separation was a result of the way that I think about the separation between platform and app, but in this case the OS dependencies can be thought of as just part of the app as well.  Perhaps they should be merged into a single "prepare the VM to run your code" build step.

One counterpoint here is that scripts are probably bad at avoiding redoing work -- I don't really want to force people to wait for an `apt-get update` to complete every time they run `vagrant-spk dev`.  On the other hand, I don't want someone trying to get a clean build to wind up with old, cached artifacts.  Perhaps this is why Docker went with the "cache the FS after each command" approach, but I dislike imperfect caching that users aren't explicitly aware of, i.e. old version of libssl package was cached and subsequent build will use the cache and still be vulnerable.

vagrant-spk
zarvox: +1

vagrant-spk
zarvox: +1

PR #14, opened 2015-06-11
Vagrantfile: switch back to vboxsf

zarvox:
This PR is blocked on https://github.com/sandstorm-io/sandstorm/pull/494 making it into a sandstorm binary release; don't merge until that one is released.

-----

A PR for spk prevents us from hitting the 4096-file-handle limit:
https://github.com/sandstorm-io/sandstorm/pull/494

So we can switch back to the default shared folder mechanism, which
has a couple of nice benefits:

1. Works on Windows at all
2. OSX users don't have to enter sudo password every time they start or stop
   the VM
3. Can map /vagrant to the same folder as /opt/app

Libvirt still does better with an explicit p9fs, which is bidirectional, as
opposed to the default synced_folder backend, which is rsync and only syncs
from host to guest.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -49,7 +49,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Then, do stack-specific and app-specific setup.
   config.vm.provision "shell", inline: "sudo bash /opt/app/.sandstorm/setup.sh"
 
-  # Shared folders are configured per-provider, since vboxsf can't handle >4096 open files,
+  # Shared folders are configured per-provider since vboxsf can't handle >4096 open files,
   # NFS requires privilege escalation every time you bring a VM up,
   # and 9p is only available on libvirt.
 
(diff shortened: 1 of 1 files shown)


paulproteus:
< 3​

vagrant-spk
```diff
@@ -101,6 +99,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     override.vm.synced_folder "..", "/opt/app", type: "9p", accessmode: "passthrough"
     override.vm.synced_folder ENV["HOME"] + "/.sandstorm", "/host-dot-sandstorm", type: "9p", accessmode: "passthrough"
+    override.vm.synced_folder "..", "/vagrant", type: "9p", accessmode: "passthrough"
```
paulproteus: I remember now how I avoided using `override`:

I would create a Ruby variable, mutate it depending on the VM provider, and then at the end, use that variable in a call to `config`.

I find the "Use a Ruby variable" way marginally cleaner, but the way you have it is also totally OK.

paulproteus:
That mak​es
You larger
Than life.

PR #16, opened 2015-06-14
Add meteor stack, built atop meteor-spk.

zarvox:


zarvox:
@paulproteus you can take this for a whirl as follows:

```bash
meteor create --example todos
cd todos
vagrant-spk setupvm meteor
vagrant-spk up
vagrant-spk init
vagrant-spk dev
# play with instance at http://local.sandstorm.io:6080
vagrant-spk pack meteor-todos.spk
```

PR #22, opened 2015-06-17
Reduce MySQL disk usage

zarvox:
This brings disk usage down to ~15MB.  I don't think we'll be able to do much
better; InnoDB requires at least 10MB for its data file path, so short of using
a patched mysqld, we can't fix this.

We could consider maintaining a patched MySQL which gets automatically pulled
in for vagrant-spk users, if that seems like a worthwhile time investment.

Worth noting: this is likely *not* a safe thing to push to existing packages with MySQL DBs.
https://dev.mysql.com/doc/refman/5.5/en/innodb-data-log-reconfiguration.html
suggests that you'd have to tell the server to completely flush the log, then
manually delete the logfiles before changing their size in the config.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -198,6 +198,14 @@ sed --in-place='' \
 sed --in-place='' \
         --expression='s/^user\t\t= mysql/#user\t\t= mysql/' \
         /etc/mysql/my.cnf
+# patch mysql conf to use smaller transaction logs to save disk space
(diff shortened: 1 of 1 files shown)


paulproteus:
If you claim you've tested this, then I approve! +1

paulproteus:
Please click "merge" whenever you want, if you have tested this.

I would love to see us grow a test suite for `vagrant-spk` but we don't have one yet!

PR #24, opened 2015-06-19
Fix command-not-found issue in setup.sh for Python stack

paulproteus:
This commit fixes an issue where the [mysqld] string is placed
before the heredoc that uses it.

Thanks to Jack Singleton for finding it.


paulproteus:
Props again to @jacksingleton .

Also this suggests to me we should start using the `installer-tests/` framework for testing that building SPKs works. Perhaps something I can look into after Open Source Bridge.

zarvox:
Wow, somehow I missed seeing this pull request until just now.  Feel free to ping me again if a PR goes silent for a long time in the future.

Thanks for finding and fixing this!

paulproteus:
Will do! I also sort of forgot about it myself.

PR #27, opened 2015-06-29
Make meteor build the bundle on the guest fs

zarvox:
On Windows, there's a path limit that Meteor/npm frequently exceed.  When we
attempt to build a bundle under /opt/app/.sandstorm/, it fails because Windows
will not allow such a long path.

This patchset makes vagrant-spk build the app in the guest OS in the vagrant
user's homedir instead.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -432,12 +432,12 @@ METEOR_WAREHOUSE_DIR="${METEOR_WAREHOUSE_DIR:-$HOME/.meteor}"
 METEOR_DEV_BUNDLE=$(dirname $(readlink -f "$METEOR_WAREHOUSE_DIR/meteor"))/dev_bundle
 
 cd /opt/app
-meteor build --directory .sandstorm
-(cd .sandstorm/bundle/programs/server && "$METEOR_DEV_BUNDLE/bin/npm" install)
+meteor build --directory /home/vagrant/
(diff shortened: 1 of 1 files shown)


zarvox:
@paulproteus Here's an untested patch; give it a shot and let me know if it works?

paulproteus:
+1. Thanks for writing this!

paulproteus:
Will test it out shortly.

zarvox:
@paulproteus ping: did you get a chance to test this out yet?

paulproteus:
Not yet; I ​hope to later today. I'm trying to think about how to reproduce
the Windows-y problems on my Linux system.

paulproteus:
I just tested this, and yes it does indeed build on the VM, which I'm +1 on.

I'd prefer it do the build in `/tmp` since these files aren't necessary to persist, and therefore it'd be nice to put these files somewhere that the operating system will remove them.

I'd also like to, in the future, place these somewhere related to the app itself, so that there's a lower chance of conflicts if people use the same VM to build multiple apps.

However, none of those are showstoppers. I'm going to merge this, since it makes things better.

Glad you worked on the patch @zarvox !

paulproteus:
For what it's worth, the `/tmp` in my `vagrant-spk` VM is not tmpfs, so I think making these large files there would work fine.

Also for what it's worth, they don't seem to get auto-cleaned-up when I `^C` the `vagrant-spk dev`.

PR #39, opened 2015-07-20
Add vagrant-vbguest configuration needed for VirtualBox 5

ndarilek:


paulproteus:
This commit is great, and the one thing I would do is add even more comments so it's clear to a maintainer of `vagrant-spk` down the road what these lines are doing here.

I'll merge this and adds some more comments on top. Thanks for taking the time to get to the bottom of this, @ndarilek , and for submitting this patch!

Issue #42, opened 2015-07-20
Purge logfiles in grain packages


JamborJan:
I realized that we get some pretty big log files in grains. I found that when looking for some information about crashing apps. So I know that log files are useful but should be purged rather quickly.

So for example we have in ```data/log/nginx``` the access log which is getting bigger and bigger. I have only in one grain I am using on my own already 2MB (Logs from March 23rd 2015 until today).

Guessing mode: 
- I guess this is something which needs to be set in the environment when packaging the app
- So I guess this needs to be done in ```vagrant-spk``` rather than sandstorm
- But I guess that I can guess wrong ...

JJ

paulproteus:
I think the thing to do is:

* Configure nginx to not store access logs at all, or at least store them in `/tmp`

* Configure nginx to route error logs to the grain log

As for existing packages, you might as well add

`rm -rf /var/log/nginx`

to `launcher.sh` I suppose!

JamborJan:
Thats basically right for nginx. I think thats not enough. What if I build an app where another component is logging a lot of stuff. I guess we should describe in the vagrant-spk docs that people packaging the app should download a grains backup it after having tested the app and check what log files are inside. For all these logfiles a garbage collector needs to be implemented to avoid that the grains get bigger and bigger.

Garbage collectors or avoiders can be:
- write log to sandstorm log (avoider, don't know how that works :-) ).
- delete entire logfile (maybe a bit too much)
- keep only the last 5 lines of the log with ```tail -5 /var/log/nginx/access.log /var/log/nginx/access.log```

Could tell me how to write something in the sandstorm log?

I can send a PR to vagrant-spk with some default files which get reduced to the last 5 lines in ```launch.sh```.

JamborJan:
I have created and tested a more generic solution for ```*.log``` files in ```/var/log```. I am using ```tail```for that.

PR #47, opened 2015-07-25
Purge logfiles in grain packages

JamborJan:
See https://github.com/sandstorm-io/vagrant-spk/issues/42

This removes all content of all ```*.log``` files in ```/var/log``` and only keeps the last 10 lines. This should prevent having big sized log files in grains which grow and grow..

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -263,6 +263,13 @@ rm -rf /var/run
 mkdir -p /var/run
 mkdir -p /var/run/mysqld
 
+# Cleanup log files
+FILES="$(find /var/log -name '*.log')"
+for f in $FILES
+do
+  tail $f | tee $f
+done
+
 # Ensure mysql tables created
 HOME=/etc/mysql /usr/bin/mysql_install_db --force
 


paulproteus:
I think the way I'd rather fix this is to adjust the `nginx` configuration to not store access log files at all. I think in a Sandstorm context, they don't serve any purpose that I can identify.

Would you be willing to make the corresponding change to the `nginx` configuration that we auto-generate in `vagrant-spk`, @JamborJan ? If not, I understand, and I can work on that myself.

Also I'd be curious if you agree/disagree with the idea that we should not bother storing these access logs at all.

JamborJan:
For the ```nginx``` access log I can understand that and the approach would be right.

My thoughts where

1) Different apps require different packages (in terms of apt-get) and software pieces which log into  ```/var/log``` and thats why a generic approach should be there to purge them. Thats why I used the approach with finding all  ```*.log``` files and processing them.

2) While an access log from ```nginx``` makes no sense at all the error log helped me to find an issue already. Thats why I wouldn't delete or send it to  ```/dev/null``` completely. Coming back to the idea of having one solution for all I did not implement one solution for the access log and one for all others. I used one for all.

paulproteus:
Hi @JamborJan ,

I note in https://github.com/sandstorm-io/vagrant-spk/blob/master/stacks/lemp/setup.sh that we do the following:

- disable access logging;

- send error logging to stderr

which results in the nginx log being part of the grain log. I think that's the best of all worlds here. Let me know if you agree, or not!

JamborJan:
Works perfect for me, we can close this.

PR #51, opened 2015-08-01
Move stack scripts out into folders

zarvox:
This should make it easier to edit, discover, and iterate on scripts.

While I was moving all the scripts, I made them all run
`set -euo pipefail` as their first action.

Tagging @paulproteus for review, if he likes.


paulproteus:
If you say you tested this with making a Meteor app, then I say:

:ship:

PR #58, opened 2015-08-20
Many php apps break without a properly created sessions directory

jparyani:


zarvox:
Yay!  Thanks for the patch.

PR #61, opened 2015-08-31
Atomically commit bundle to ~/.sandstorm/caches/

### LINKED ISSUE #60: In the case of partially-downloaded sandstorm bundle tar.xz, re-download it rather than fail

zarvox:
This will cause redownload rather than failure if a download fails partway
through.

Fixes #60.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -123,7 +123,8 @@ curl https://install.sandstorm.io/ > /host-dot-sandstorm/caches/install.sh
 SANDSTORM_CURRENT_VERSION=$(curl -fs "https://install.sandstorm.io/dev?from=0&type=install")
 SANDSTORM_PACKAGE="sandstorm-$SANDSTORM_CURRENT_VERSION.tar.xz"
 if [[ ! -f /host-dot-sandstorm/caches/$SANDSTORM_PACKAGE ]] ; then
-    curl --output "/host-dot-sandstorm/caches/$SANDSTORM_PACKAGE" "https://dl.sandstorm.io/$SANDSTORM_PACKAGE"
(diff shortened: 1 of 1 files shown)


zarvox:
@paulproteus review and merge? :)

vagrant-spk
```diff
@@ -123,7 +123,8 @@ curl https://install.sandstorm.io/ > /host-dot-sandstorm/caches/install.sh
 SANDSTORM_CURRENT_VERSION=$(curl -fs "https://install.sandstorm.io/dev?from=0&type=install")
 SANDSTORM_PACKAGE="sandstorm-$SANDSTORM_CURRENT_VERSION.tar.xz"
 if [[ ! -f /host-dot-sandstorm/caches/$SANDSTORM_PACKAGE ]] ; then
-    curl --output "/host-dot-sandstorm/caches/$SANDSTORM_PACKAGE" "https://dl.sandstorm.io/$SANDSTORM_PACKAGE"
+    curl --output "/host-dot-sandstorm/caches/$SANDSTORM_PACKAGE.partial" "https://dl.sandstorm.io/$SANDSTORM_PACKAGE"
```
paulproteus: Yay thanks!

In theory mumble mktemp mumble, but ~/.sandstorm/caches/ is a safe place, so I think mktemp isn't important.

zarvox: A fair point.  Sadly, we already have the limitation that you can only be running one `vagrant-spk` VM at a time (or at least only port forward one of the VMs to port 6080 at a time...) so I suspect it'll be okay and this way we don't leak disk space in the case of failed downloads.

paulproteus:
+1 to this self-merge

PR #62, opened 2015-08-31
Add the "publish" command

jparyani:
I couldn't think of a better way than to "copy" the spk, since it can be
anywhere on the host filesystem.


zarvox:
Awesome, thanks for this!

PR #66, opened 2015-09-08
Make better VM names based on the dirname containing the .sandstorm dir

zarvox:


vagrant-spk
```diff
@@ -113,7 +117,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider :libvirt do |libvirt, override|
     libvirt.cpus = cpus
     libvirt.memory = assign_ram_mb
-    libvirt.random_hostname = true
+    libvirt.default_prefix = PROJECT_BASENAME + "_sandstorm_#{Time.now.utc.to_i}"
```
paulproteus: +0.9, I prefer randomness but then again there's nothing wrong with this in practice for this use.

paulproteus: And @jparyani prefers timestamps.

vagrant-spk
```diff
@@ -105,6 +108,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider :virtualbox do |vb, override|
     vb.cpus = cpus
     vb.memory = assign_ram_mb
+    vb.name = PROJECT_BASENAME + "_sandstorm"
```
paulproteus: Let me test this now, to see it causes duplicates.

IMHO we should do the same as for `libvirt`.

I propose you calculate that string in a variable, rather than do it once per `config.vm.provider`.

paulproteus: Yes, it does in fact conflict.

```
==> default: `vagrant box update` to update.
A VirtualBox machine with the name 'same_sandstorm' already exists.
Please use another name or delete the machine with the existing
name, and try again.
Traceback (most recent call last):
  File "/usr/local/bin/vagrant-spk", line 378, in <module>
    main()
  File "/usr/local/bin/vagrant-spk", line 375, in main
    operation(args)
  File "/usr/local/bin/vagrant-spk", line 291, in bring_up_vm
    call_vagrant_command(sandstorm_dir, "up")
  File "/usr/local/bin/vagrant-spk", line 188, in call_vagrant_command
    return subprocess.check_call(command, cwd=sandstorm_dir)
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['vagrant', 'up']' returned non-zero exit status 1
```

zarvox: Darn.

I'm mildly concerned that setting vm.name to a non-random value will attempt to change the name every time you run a vagrant command, which would be annoying.  I guess it's not a particularly-bad behavior though, just slightly surprising.

zarvox:
All right, give it another whirl.

vagrant-spk
```diff
@@ -105,6 +110,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider :virtualbox do |vb, override|
     vb.cpus = cpus
     vb.memory = assign_ram_mb
+    vb.name = VM_NAME
```
paulproteus: I've tested this with VirtualBox and nothing too terrible seems to happen! Yay.

paulproteus:
merged! Many thanks. I'm pretty excited about this "small' change.

PR #69, opened 2015-09-23
use meteor-spk's start.js

dwrensha:
This is dependent on https://github.com/sandstorm-io/meteor-spk/pull/10 and a new release of meteor-spk being pushed. (0.1.5 is the current release)

### DIFF
--- a/stacks/meteor/setup.sh
+++ b/stacks/meteor/setup.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 cd /opt/
 
-PACKAGE=meteor-spk-0.1.4
+PACKAGE=meteor-spk-0.1.6
 PACKAGE_FILENAME="$PACKAGE.tar.xz"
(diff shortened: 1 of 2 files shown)


zarvox:
LGTM.

Do we want to also update the Meteor version we're downloading/caching to 1.2?

kentonv:
Err, oops, you're going to want to update that to 0.1.7 due to my goof in pushing 0.1.6 without actually including David's change...

stacks/meteor/setup.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 cd /opt/
 
-PACKAGE=meteor-spk-0.1.4
+PACKAGE=meteor-spk-0.1.6
```
zarvox: @kentonv says this needs to be 0.1.7 because he flubbed the first release - I'll take care of this.

PR #70, opened 2015-10-02
nginx: by default, don't limit max body size

zarvox:
nginx will limit this to 1MB by default, which is pretty small for a lot of
apps that allow e.g. photo or media uploads.


zarvox:
@jparyani @dwrensha You probably want to apply similar patches to any apps you packaged and release updates.  In particular, Lychee is known to be affected.

paulproteus:
​+1

in future would love to see nginx conf in .vagrant/ but I don't need to let
the perfect be the enemy of the good.

Issue #53, opened 2015-08-05
Provide more useful --help


paulproteus:
@pwais just did `wipe` and thought he was going to remove his VM but actually removed his scripts. Apparently it's _mostly_ OK this time, but anyway he would like more `--help`!

pwais:
Yep, I'm a doofus.  "Wipe" sounded like "wipe the VM" not "wipe the current directory" to me, and "destroy" sounded more like "destroy the directory" to me. I should have just read the code X_X .

Don't mean to toot my own horn, but here's an example (off the top of my head) of using OptionParser for better `--help` text: https://gist.github.com/pwais/b8975650fa39a7750d43   Though I'm not sure if you're looking to target Python versions that don't have OptionParser.

paulproteus:
You're not a doofus. Thanks for showing me the issues you have been having
so we can continue to make things better.

paulproteus:
I believe that #71 closes this. Thanks @pwais for working on it! (And we removed the pain-inducing `wipe`!)

PR #71, opened 2015-10-05
More useful --help text.  Ref issue #53

pwais:
Some more useful text for `--help`.  Aims to close issue #53 .  FWIW I see that `--wipe` was removed in 4ef5a210a7ee2a42b4a8b0abd8924f7d037e5def  (thanks :) )

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -327,8 +327,7 @@ def pack(args):
     print("package produced at {}".format(output_spk))
 
 def publish(args):
-    'Sadly, we have to copy the spk into the VM since it could be anywhere on the host filesystem'
-
+    # Sadly, we have to copy the spk into the VM since it could be anywhere on the host filesystem
     spk = args.command_specific_args[0]
     sandstorm_dir = os.path.join(args.work_directory, ".sandstorm")
     spk_basename = os.path.basename(spk)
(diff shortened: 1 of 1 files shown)


pwais:
Oh, tested (and should work) for 80-col terminal.

vagrant-spk
```diff
@@ -356,24 +355,52 @@ def ssh(args):
...
-            'global-status': global_status,
-            'ssh': ssh,
-    }
-    parser = argparse.ArgumentParser(prog=sys.argv[0])
-    parser.add_argument("command", choices=sorted(operations.keys()))
+    operations = (
+        ('setupvm', setup_vm,
+            "Set up a fresh app environment in \n"
+            "  --work-directory (using the provided `stack` \n"
+            "  command arg).  May also init credentials in \n"
```
zarvox: Adding keys to the keyring actually happens over in `init` - can you move that comment over there, and make it unconditional?

zarvox:
This is great!  Thanks a ton for making this better!

I'm happy to merge this as-is and make any changes I requested myself, or you can if you like.  Let me know!

In the fullness of time, it might make sense to make the command structure into a class or at least a named tuple, but I can also take that on if you prefer.

vagrant-spk
```diff
@@ -356,24 +355,52 @@ def ssh(args):
...
+            "Delete the Vagrant VM (but leave \n"
+            "  --work-directory untouched)."),
+        ('global-status', global_status,
+            "Show the state of all active \n"
+            "  Vagrant VMs on this host."),
+        ('ssh', ssh,
+            "SSH into the Vagrant VM."),
+    )
+    op_to_func = dict((cmd, f) for cmd, f, helptext in operations)
+    ops_helptext = '\n'.join(cmd + ': ' + helptext for cmd, f, helptext in operations)
```
zarvox: I think I'd be happier if the help text and the choices list were sorted in the same order, alphabetically.  It's a little bit sad that we lose the "lifecycle ordering" above, but I think consistency is more valuable, and most people will probably use `--help` to remember what we call specific subcommands, rather than to learn the high-level lifecycle concepts, which I think are laid out in more detail in the sandstorm docs.

paulproteus: +1 to Drew's thoughtful feedback.

​Paul, if you're willing to sign our CLA, that'd be helpful. It may not be
required for such a small code change, but it would help smooth the path
for future ones.

https://sandstorm.io/cla/ is the URL for it. This is the same CLA as the
Apache Software Foundation uses, if that's useful information.

I think it's OK to merge pre-CLA but I thought now would be a good time to
ask.

vagrant-spk
pwais: hmm, `ensure_host_sandstorm_folder_exists()` only happens on `up` and `setupvm`; that's what i meant about 'credentials'.  I'll update the comment

vagrant-spk
pwais: Removed the `sorted()` to make the ordering consistent.  I actually tried out `--help` before really looking at the docs... I was looking for initial commands to run.

pwais:
✓ updated `setupvm` note about credentials / keys
✓ made command arg ordering match helptext (lifecycle order)
✓ signed Sandstorm CLA

I'm not sure I really grok what is wanted out of a class structure or namedtuple wrapping the help stuff. Also unclear if lifecycle order for the helptext was undesirable or having the arg list order consistent with helptext suffices.  I can try to help refine from here, but TBH I know more about what help I was looking for than the expected maintenance future for this script :)

zarvox:
Lifecycle order consistently is fine.  This is plenty to make me happy for now.

Class structure would be just making a class with a name in place of the anonymous 3-tuples of (`command_name`, `command_action`, `help_text`).  Then, in the places they're used, you could pull out individual keys rather than having to care about/extract all the fields into unused variables.  That way we could write (for instance):

```python
    op_to_func = dict((op.name, op.action) for op in operations)
```

rather than:

```python
    op_to_func = dict((cmd, f) for cmd, f, helptext in operations)
```

which references helptext, even though it only wants to discard it.  If we wanted to add another field to operations (say, longer help text if you called `vagrant-spk help <subcommand>`), then we'd have to change the code as merged to throw away that fourth member, whereas the suggested structuring wouldn't have to change.

---

I guess if you looked at the `--help` output to see how to get started, that suggests my intuition about the docs isn't necessarily correct for everyone. :)  We may revisit this in the future, but for now, thanks, and I'm excited to get better help text in the hands of users!

pwais:
Thanks for the merge @zarvox !

Ah, I see what you were getting at with namedtuples now.  Also agree extended helptext / help per-command could be great; I find that sort of feature especially helpful for really complex tools like  `gcloud`.  Glad things seem to suffice for now, and I imagine these issues will come naturally to whoever touches this code next.

PR #74, opened 2015-10-10
Quieter output for meteor stack

paulproteus:
Hi @zarvox,

I think I'm going to self-merge this because I have manually tested it, and it works OK for Meteor apps, and I have a goal of making sure the hackathon people tomorrow have the best experience possible.

If you're +1 on these changes I can make them to other stacks, too, and/or you can if you prefer to for some reason.

Semi-apologies for the self-merge.

### DIFF
--- a/stacks/meteor/setup.sh
+++ b/stacks/meteor/setup.sh
@@ -18,7 +18,7 @@ fi
 tar xf "$CACHE_TARGET"
 
 # Create symlink so we can rely on the path /opt/meteor-spk
-ln -s "${PACKAGE}" meteor-spk
+if [ ! -e meteor-spk ] ; then ln -s "${PACKAGE}" meteor-spk ; fi
 
 # Add bash, and its dependencies, so they get mapped into the image.
 # Bash runs the launcher script.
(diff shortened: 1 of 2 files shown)


zarvox:
Sorry, but I'm -1 on parts of this.

While I agree with the general idea of not making things overly verbose about implementation details, I reject the notion that this means we should silence everything.  It's particularly helpful to be able to distinguish "making progress" from "hanging", which this changeset regresses on.  First-time users are going to have to download the largish Sandstorm tarball and meteor-spk tarballs on the first `vagrant-spk up`, and if their internet connection isn't amazing, it looks like the script has gotten stuck while it performs that download.  We need to find a happy medium between silent and excessively verbose.

When a command is likely to take longer than a couple of seconds to complete, we should indicate what is happening, so that users are not sitting wondering if the program has gotten stuck or failed.

Perhaps the action item here is to print some messages about the forward progress that the setup script is making, or that a particular step may take a while.  Additionally, while I'd like to think that our tools will work perfectly every time, a handful of progress messages to paste can help us debug what is succeeding and failing if something goes wrong for a user in the setup flow.

paulproteus:
+1 to progress messages that show people that something is basically happening. Let me see if I can introduce those in a new pull request.

paulproteus:
i,i I had this unsent email draft in my email:

curl --silent --show-error means that an error will be printed out if
there's an actionable error. I wonder if we agree more than we think. Let's
discuss more IRL if it can wait until Monday. Feel free to revert if you
think that's best until then.


Anyway, thanks for your help improving this, Drew!

PR #77, opened 2015-10-14
lemp: pass environment variables on to php-fpm

zarvox:
This allows us to set SANDSTORM=1 in the default environment block generated by
`spk init` and then call getenv("SANDSTORM") in PHP apps to see if they're
being run under Sandstorm or not.

### DIFF
--- a/stacks/lemp/setup.sh
+++ b/stacks/lemp/setup.sh
@@ -49,6 +49,11 @@ sed --in-place='' \
 sed --in-place='' \
         --expression='s/^pid =/#pid =/' \
         /etc/php5/fpm/php-fpm.conf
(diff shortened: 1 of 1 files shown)


paulproteus:
I'd vaguely prefer to whitelist specific env vars, but I don't see a way to do that in the PHP docs, so +1, merging.

zarvox:
Yeah.  You can explicitly set `env[SANDSTORM] = $SANDSTORM` in the same file, but that only works if you haven't already cleared the environment...

...and the `sandstorm-pkgdef.capnp` specifies a hermetic environment to start from, so anything else present is added by your sandstorm-specific scripts anyway.

paulproteus:
+1​

PR #76, opened 2015-10-12
Don't set ALLOW_DEV_ACCOUNTS; the Sandstorm install.sh does this

zarvox:

PR #79, opened 2015-10-17
Need to restart so the new BIND_IP takes effect.

dwrensha:
After #77, on an initial `vagrant-spk up` Sandstorm does not become available on local.sandstorm.io:6080. This patch fixes the problem.


dwrensha:
Er, the problem was #76, I mean.

paulproteus:
In _theory_ we don't need to always restart, just if we changed the `BIND_IP`, but life is short. I'm going to merge this.

PR #81, opened 2015-10-20
Add an experimental auto-packaging command

paulproteus:
I have tested that this works properly on the Meteor clock. For now, my
main target user is @neynah -- my goal is to set up a process where she
builds sample versions of packages for people so they understand how
cool it would be if they created a "real" package.

Therefore @zarvox I am hopeful that we can have lower code quality behind
the VAGRANT_SPK_EXPERIMENTAL flag. I promise not to change how existing,
non-experimental things work via changes that are only available in
EXPERIMENTAL mode.

FWIW, in the near/long run, this auto mode might be a good idea for real
Sandstorm app authors. I know I would use it, and I actually understand
Sandstorm.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -29,6 +29,7 @@ import subprocess
 import sys
 
 PWD = os.getcwd()
+ENABLE_EXPERIMENTAL = False
 CODE_DIR = os.path.dirname(os.path.realpath(sys.argv[0]))
 VAGRANTFILE_CONTENTS = r"""# -*- mode: ruby -*-
 # vi: set ft=ruby :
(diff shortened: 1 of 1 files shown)


paulproteus:
Requesting an OK on merge from @zarvox .

Next steps on smoothing out the app author pipeline:

* Create a Sandstorm app that can generate a simulated app market UI, containing this app.

* Automate uploading the app to Oasis so that it's available on demo.sandstorm.io with this (obscure) app ID (if possible)

* Decide which parts of this @neynah actually cares about.

paulproteus:
To reproduce my success:

```
mkdir -p ~/projects/sandstorm-packaging-tutorial  
cd ~/projects/sandstorm-packaging-tutorial
meteor create --example clock
cd clock
VAGRANT_SPK_EXPERIMENTAL=Y vagrant-spk auto meteor
```

Now check out ~/projects/sandstorm-packaging-tutorial/app.spk

neynah:
Wow! Awesome!

zarvox:
Feel free to merge (nice work hiding behind a feature flag).

It's kinda weird that this calls `setupvm` unconditionally.  It gives me pause to running it twice on the same source tree, and also means that anything that happens in setup/build/launcher is limited to be purely the defaults in vagrant-spk for that stack.  Which is rather frequently not what we ship.

I'd be happier if it only bothered creating the `.sandstorm` if none existed already, which would have the added benefit of users being able to use this with e.g. repos that they put on GitHub that might have some changes in setup.sh/build.sh.

paulproteus:
+1, I want setupvm etc. to be conditional. Just wanted to make a v0.​

PR #89, opened 2015-10-27
Add node stack

mnutt:
Unfortunately there's not one single standardized way to deploy node apps, so this stack attempts to use the minimal reasonable defaults to get someone up and running with a node app:

* Installs the latest stable node.js (4.x)
* `npm install` and `npm start`

I think it could probably be argued that it's too minimal and the diy stack suffices. The main value is really just using nodesource to get a reasonably up to date node in the VM.

### DIFF
--- a/stacks/node/launcher.sh
+++ b/stacks/node/launcher.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -euo pipefail
+# This script is run every time an instance of our app - aka grain - starts up.
+# This is the entry point for your application both when a grain is first launched
+# and when a grain resumes after being previously shut down.
+#
+# This script is responsible for launching everything your app needs to run.  The
+# thing it should do *last* is:
+#
+#   * Start a process in the foreground listening on port 8000 for HTTP requests.
+#
+# This is how you indicate to the platform that your application is up and
+# ready to receive requests.  Often, this will be something like nginx serving
+# static files and reverse proxying for some other dynamic backend service.
+#
+# Other things you probably want to do in this script include:
+#
+#   * Building folder structures in /var.  /var is the only non-tmpfs folder
+#     mounted read-write in the sandbox, and when a grain is first launched, it
+#     will start out empty.  It will persist between runs of the same grain, but
+#     be unique per app instance.  That is, two instances of the same app have
+#     separate instances of /var.
+#   * Preparing a database and running migrations.  As your package changes
+#     over time and you release updates, you will need to deal with migrating
+#     data from previous schema versions to new ones, since users should not have
+#     to think about such things.
+#   * Launching other daemons your app needs (e.g. mysqld, redis-server, etc.)
+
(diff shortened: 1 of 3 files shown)


paulproteus:
Very exciting!

Some discussion going on, on IRC - logs here: https://botbot.me/freenode/sandstorm/2015-10-27/?msg=52811978&page=1

mnutt:
Ok, updated to inline the node install and make express.js (the most common node framework) more likely to work.

stacks/node/launcher.sh
```diff
@@ -0,0 +1,36 @@
...
+#     data from previous schema versions to new ones, since users should not have
+#     to think about such things.
+#   * Launching other daemons your app needs (e.g. mysqld, redis-server, etc.)
+
+# By default, this script does nothing.  You'll have to modify it as
+# appropriate for your application.
+cd /opt/app
+
+# express.js has fewer problems when HOME is defined
+export HOME=/opt/app
```
paulproteus: This should probably be `/var` since `/opt/app` isn't writeable.

It could even be `/tmp` if you don't mind it being erased between runs of the app! (I think honestly this is probably correct, but you might want to leave a note in the shell comments if you go that route.)

stacks/node/setup.sh
```diff
@@ -0,0 +1,87 @@
...
+sckuXINIU3DFWzZGr0QrqkuE/jyr7FXeUJj9B7cLo+s/TXo+RaVfi3kOc9BoxIvy
+/qiNGs/TKy2/Ujqp/affmIMoMXSozKmga81JSwkADO1JMgUy6dApXz9kP4EE3g==
+=CLGF
+-----END PGP PUBLIC KEY BLOCK-----
+" | apt-key add -
+
+apt-get update
+
+# Actually install node
+apt-get install -qq nodejs git-core g++
```
paulproteus: +1

paulproteus:
I like this already.

Some questions:

* Is there a sample node/express app I can package this way? I'd love to make a copy-pasta version of https://docs.sandstorm.io/en/latest/vagrant-spk/packaging-tutorial-meteor/ for a nodejs/express app, plus then I could verify the instructions are plausible.

Maybe just that one question actually!

paulproteus:
Hi @mnutt !

Is there a git repo you can provide to a package that works with this, so I can test it?

If not, I can try to review it blind, but it seems wisest to have something to test it with (-:

Thanks!

paulproteus:
Life is short. I'll merge this.

After testing it, I'll write docs.

mnutt:
Sorry for not following up on this! After you asked, I then looked around at a bunch of open source node projects, and almost all of them ran their code a different way. This way seems like a pretty reasonable way to package a node app to me, but it was disheartening that it's probably not going to be drop-in for a new user trying to package an existing app.

Let me know if I can help with documentation here.

PR #95, opened 2015-10-31
meteor stack: bump meteor-spk version to 0.1.8

zarvox:
Fixes a bug in migrations.  We should notify app authors to bump their versions
and `vagrant-spk destroy` and `vagrant-spk up` before their next release.


zarvox:
@paulproteus I'm self-merging for now, but a release/bugs against wekan/any other apps that use the meteor stack would probably be a nice followup.

paulproteus:
+1​

PR #97, opened 2015-11-04
Newline sanity, per @jparyani

paulproteus:


paulproteus:
Self-merging as purely experimental

Thanks to @jparyani for the feedback!

PR #98, opened 2015-11-04
Append not clobber

paulproteus:
Once again thanks @jparyani


paulproteus:
Self-merging because experimental

PR #99, opened 2015-11-04
vagrant-spk auto: Automatically upgrade HTTP Google Fonts to HTTPS

paulproteus:
This helps Meteor apps fully load when packaged and run on HTTPS
Sandstorm servers.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -25,6 +25,7 @@ from __future__ import print_function
 import argparse
 import os
 import glob
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


paulproteus:
Self-merging since this only affects experimental code.

ocdtrekkie:
I have a question: Don't we want to not allow Google fonts? Or, to be more useful, auto-download and include them?

paulproteus:
+1 to that. This is just a short-term hack to get this app working, but I
do want that isolation.​

PR #104, opened 2015-11-12
Add "verify" subcommand.

zarvox:
Also run `spk verify --details` when calling `spk pack` since users are almost
invariably going to want to see that output to sanity-check their metadata.

@paulproteus can I talk you into writing the appropriate docs for this? :)

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -487,10 +487,23 @@ def pack(args):
     call_vagrant_command(sandstorm_dir, "ssh", "-c", " && ".join([
         "cd /opt/app/.sandstorm/",
         "spk pack --keyring=/host-dot-sandstorm/sandstorm-keyring --pkg-def=/opt/app/.sandstorm/sandstorm-pkgdef.capnp:pkgdef /home/vagrant/sandstorm-package.spk",
+        "spk verify --details /home/vagrant/sandstorm-package.spk",
         "mv /home/vagrant/sandstorm-package.spk /opt/app/sandstorm-package.spk"
     ]))
     os.rename("sandstorm-package.spk", output_spk)
     print("package produced at {}".format(output_spk))
+
+def verify(args):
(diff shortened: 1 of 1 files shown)


paulproteus:
Just thinking about the docs now.

The `vagrant-spk verify` command does `spk verify --details` and would not act properly in the case of `vagrant-spk verify --details filename.spk`, I _think_

Given that, any chance that @zarvox you would also (at the same time) adjust `spk` to be `--details` by default?

paulproteus:
And/or make this one _not_ do `--details` by default, and/or accept `--details` optionally.

My issue is that I don't want to to say in the docs:

> Run `vagrant-spk verify filename.spk` (but if you use the raw SPK tools, `spk verify --details filename.spk`)

paulproteus:
i,i documentation-driven development

zarvox:
Maybe, but I'm definitely punting until at least blur day

paulproteus:
I can live with revisiting this mid-day Friday

zarvox:
I'm +1 on `spk verify` showing details by default; I don't think anything depends on its existing behavior of printing just the app ID

zarvox:
Okay, I lied about putting this off until blur day: https://github.com/sandstorm-io/sandstorm/pull/1155

We can drop the `--details` from the `vagrant-spk` command once a Sandstorm release prints details by default for long enough.  We can drop the `--details` from the docs as soon as the PR linked here and this PR are merged.

zarvox:
Okay, we can merge this now.  After the next sandstorm release, the docs can say `{vagrant-,}spk verify package.spk` consistently.

paulproteus:
I'm going to merge this now, even though there's been no docs update. I will file a bug requesting the docs update.

paulproteus:
https://github.com/sandstorm-io/sandstorm/issues/1370

here's the bug!

PR #105, opened 2015-12-14
Pin to version 8.2.1 of debian/jessie64, for vboxsf support

paulproteus:
Currently doing the tutorial now, to make sure this works.

Also it sure is a shame we didn't catch this earlier. This box got released 19 days ago: https://atlas.hashicorp.com/debian/boxes/jessie64


paulproteus:
I won't merge this, but it's a useful workaround for those who need it right now.

PR #107, opened 2015-12-15
If the base box depends on debian/jessie64, say how to proceed

paulproteus:
@zarvox please review/merge soon, and once this is in, I will cut a release and write release notes.

Please read the strings carefully; I'm happy to take feedback on them, but I am hoping to get this done soon so I want as low-latency and high-confidence a review as possible. I'm willing to make string changes now; less excited about doing so as time moves forward.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -206,7 +206,7 @@ def ensure_working_vboxsf_in_base_box(sandstorm_dir):
     def test_vagrantfile_pins_version():
         with open(os.path.join(sandstorm_dir, 'Vagrantfile'), 'r') as fd:
             for line in fd:
-                if line.strip().startswith('config.vm.box_version '):
+                if line.strip().startswith('config.vm.box_version'):
                     return True
         return False
     vagrantfile_pins_version = test_vagrantfile_pins_version()
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -191,6 +191,72 @@ def call_vagrant_command(sandstorm_dir, *command_args):
     print("Calling {} in {}".format(" ".join(["'{}'".format(arg) for arg in command]), sandstorm_dir))
     return subprocess.check_call(command, cwd=sandstorm_dir)
 
+def ensure_working_vboxsf_in_base_box(sandstorm_dir):
+    # If the .sandstorm/Vagrantfile refers to debian/jessie64, and does not pin it to a particular
+    # version, and the user has version 8.2.2 available on their system, then we abort the
+    # "vagrant-spk up" process because it will not work.
+    def test_vagrantfile_refers_to_jessie64():
+        with open(os.path.join(sandstorm_dir, 'Vagrantfile'), 'r') as fd:
+            for line in fd:
+                if line.strip().split() == 'config.vm.box = "debian/jessie64"'.split():
+                    return True
+        return False
+    vagrantfile_refers_to_jessie64 = test_vagrantfile_refers_to_jessie64()
+
+    def test_vagrantfile_pins_version():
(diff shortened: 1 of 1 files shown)


vagrant-spk
```diff
@@ -191,6 +191,72 @@ def call_vagrant_command(sandstorm_dir, *command_args):
     print("Calling {} in {}".format(" ".join(["'{}'".format(arg) for arg in command]), sandstorm_dir))
     return subprocess.check_call(command, cwd=sandstorm_dir)
 
+def ensure_working_vboxsf_in_base_box(sandstorm_dir):
+    # If the .sandstorm/Vagrantfile refers to debian/jessie64, and does not pin it to a particular
+    # version, and the user has version 8.2.2 available on their system, then we abort the
```
zarvox: should this check for 8.2.2 or above, or only specific known bad versions?

vagrant-spk
```diff
@@ -191,6 +191,72 @@ def call_vagrant_command(sandstorm_dir, *command_args):
...
+            for line in fd:
+                if line.strip().split() == 'config.vm.box = "debian/jessie64"'.split():
+                    return True
+        return False
+    vagrantfile_refers_to_jessie64 = test_vagrantfile_refers_to_jessie64()
+
+    def test_vagrantfile_pins_version():
+        with open(os.path.join(sandstorm_dir, 'Vagrantfile'), 'r') as fd:
+            for line in fd:
+                if line.strip().startswith('config.vm.box_version '):
```
zarvox: Probably don't want to assume that there's a whitespace after this - it could be followed immediately by an `=`, for instance.

vagrant-spk
paulproteus: I'd like to start by just checking for known-bad versions; hopefully 8.2.3 can revert this, though I doubt it, but let's hope for that for now IMHO.

vagrant-spk
paulproteus: +1

paulproteus: Fixed.

It hadn't occurred to me that anyone would write

`var=value`

since

`var = value`

looks _so_ much nicer. But you're totally right here. (-:

vagrant-spk
```diff
@@ -191,6 +191,72 @@ def call_vagrant_command(sandstorm_dir, *command_args):
...
     return subprocess.check_call(command, cwd=sandstorm_dir)
 
+def ensure_working_vboxsf_in_base_box(sandstorm_dir):
+    # If the .sandstorm/Vagrantfile refers to debian/jessie64, and does not pin it to a particular
+    # version, and the user has version 8.2.2 available on their system, then we abort the
+    # "vagrant-spk up" process because it will not work.
+    def test_vagrantfile_refers_to_jessie64():
+        with open(os.path.join(sandstorm_dir, 'Vagrantfile'), 'r') as fd:
+            for line in fd:
+                if line.strip().split() == 'config.vm.box = "debian/jessie64"'.split():
```
zarvox: This line may also differ in whitespace presence/absence, though we generally generate this line, so it's probably not modified by anyone in the common case.

zarvox:
Strings are fine.  You can choose to try to make this robust to whitespace differences or not.

Merge when you're satisfied.

vagrant-spk
paulproteus: The `split()` is supposed to handle some variation, but I can live with people hand-modifying this line losing out, fwiw.

```
>>> 'var = value'.split()
['var', '=', 'value']
>>> 'var    = value'.split()
['var', '=', 'value']
>>> 'var= value'.split()
['var=', 'value']
```

So I guess it's true that I don't catch all variations.

I'll live with this for now. Thank you for noticing it, though; I hadn't thought clearly enough about the variations.

PR #112, opened 2016-01-04
Use semicolon as PHP config comment marker

### LINKED ISSUE #111: '#' comments are deprecated in PHP configuration files
Spun up a new LEMP stack today and got the following error in my app logs:

2016/01/04 17:03:01 [error] 19#0: *2 FastCGI sent in stderr: "PHP message: Fatal error: Comments starting with '#' are deprecated in Unknown on line 1" while reading response header from upstream, client: 127.0.0.1, server: localhost, request: "GET / HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "ceca7e9216ba13f1d457915e864631ba.local.sandstorm.io:6080"

A bit of Googling led me here:

http://stackoverflow.com/questions/14074101/getting-comments-starting-with-are-deprecated-message-via-cli

which suggests that '#' comments in PHP configs are deprecated in favor of ';'.

Not sure whether this will work, but I've updated all of my shell scripts to use ; comments and will regenerate the VM. I'll report back on whether it fixes the issue, but thought I'd at least log the error here.

paulproteus:
Fixes #111

Thanks @ndarilek for finding this and filing it.

### DIFF
--- a/stacks/lemp/setup.sh
+++ b/stacks/lemp/setup.sh
@@ -40,14 +40,14 @@ systemctl disable php5-fpm
 systemctl disable mysql
 # patch /etc/php5/fpm/pool.d/www.conf to not change uid/gid to www-data
 sed --in-place='' \
-        --expression='s/^listen.owner = www-data/#listen.owner = www-data/' \
-        --expression='s/^listen.group = www-data/#listen.group = www-data/' \
-        --expression='s/^user = www-data/#user = www-data/' \
-        --expression='s/^group = www-data/#group = www-data/' \
+        --expression='s/^listen.owner = www-data/;listen.owner = www-data/' \
+        --expression='s/^listen.group = www-data/;listen.group = www-data/' \
+        --expression='s/^user = www-data/;user = www-data/' \
+        --expression='s/^group = www-data/;group = www-data/' \
         /etc/php5/fpm/pool.d/www.conf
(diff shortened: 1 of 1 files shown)


paulproteus:
@ndarilek hoping for your +1 on this if you are willing.

@zarvox happy to await your review if you're around today.

ndarilek:
I don't know that the changes to /etc/mysql configuration files are needed/valid here since those aren't PHP. I won't be able to test those since the app I'm packaging doesn't require MySQL so I removed those bits.

I'll reply with other details later.

stacks/lemp/setup.sh
```diff
@@ -56,7 +56,7 @@ sed --in-place='' \
         /etc/php5/fpm/pool.d/www.conf
 # patch mysql conf to not change uid
 sed --in-place='' \
-        --expression='s/^user\t\t= mysql/#user\t\t= mysql/' \
+        --expression='s/^user\t\t= mysql/;user\t\t= mysql/' \
```
paulproteus: Yeah I'll undo this, one sec.

ndarilek:
I don't know that the changes to /etc/mysql configuration files are needed/valid here since those aren't PHP. I won't be able to test those since the app I'm packaging doesn't require MySQL so I removed those bits.

I'll reply with other details later.

zarvox:
Looks great.  Thanks @ndarilek for the report and @paulproteus for the fix!

PR #123, opened 2016-01-30
Install git by default in uwsgi stack

### LINKED ISSUE #119: Discussion topic: Should we provide git in the default base box for Python?
Yesterday @phildini and I were working on a Python package, and his app's `requirements.txt` listed a dependency pinned to a specific git commit.

Eventually the `pip install -r requirements.txt` failed with an error that `git` could not be found.

So I wanted to raise this discussion topic: Should we provide git in the default (or perhaps just `uwsgi`) base box?

- Upside: `pip install` (which we do automatically run on behalf of the user) would work more frequently.

- Downside: The box would take longer to set up.

I gently vote for yes, but I wanted to get @zarvox 's opinion on this.

zarvox:
Pip packages can reference git remotes, which is actually done a decent bit in
the Python community.  This should help things be more likely to work out of
the box, and git doesn't pull in too many dependencies, so it shouldn't be too
painful for users that don't need it.

Closes #119

### DIFF
--- a/stacks/uwsgi/setup.sh
+++ b/stacks/uwsgi/setup.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
-apt-get install -y nginx mysql-server libmysqlclient-dev uwsgi uwsgi-plugin-python build-essential python-dev python-virtualenv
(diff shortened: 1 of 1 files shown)


paulproteus:
+1 thanks for doing this!

PR #124, opened 2016-01-30
Disable mounting of /vagrant

zarvox:
Libvirt refuses to mount two guest folders from the same host folder.  Since we
don't use `/vagrant` anywhere at all; preferring `/opt/app`, we can just
canonicalize on `/opt/app` and disable `/vagrant` entirely.


zarvox:
@paulproteus Would love to hear your take on this.

paulproteus:
+1, I don't care about /vagrant​

We should document it as a new change, but luckily this doesn't change any
existing apps! So mega +1

zarvox:
Cool, I'll let you mention it in the release notes for the next release. :)

paulproteus:
b

PR #125, opened 2016-01-31
Add `service-config` for stacks and use in static, lemp, and uwsgi stacks

### LINKED ISSUE #116: Idea: For platform stacks that use nginx, move nginx conf into /opt/app/
Purpose of this issue:

- Propose an idea, and ask @zarvox and others if it is a good idea, and request other better ideas in the process.

Plan:

- I hope people discuss this idea here, at least saying +1 or -1, and preferably suggesting other ideas or specific criticisms.

- If there seems to be consensus for it, then I'll write, "Seems like we all think it's a good idea. Great!", and then I'll close this, and then I'll file a bug actually requesting the change. If there isn't consensus for it, or there is indeed consensus it's a bad idea, then I will close this.

- My timeline is I'd like to decide yes/no on this by next Friday (Fri Jan 29). No massive urgency and I can accept some delay; just leave a comment requesting it.

Status quo:

- `setup.sh` mutates a global set of nginx configuration files.
- This means that when one wants to modify the nginx configuration, one must `cd .sandstorm ; vagrant provision ; cd ..`
- Additionally @nicosomb just ran into a problem where he was using `$1` as part of the nginx configuration (to mirror https://www.nginx.com/resources/wiki/start/topics/recipes/symfony/ ) and ran into a problem: `==> default: /opt/app/.sandstorm/setup.sh: line 9: $1: unbound variable` ; I _think_ bash heredocs support variable expansion, which neither he nor I expected in this case.

Proposal:

- For the platform stacks where we use nginx (`lemp` `uwsgi` `static`) (but using the filenames from `lemp` as an example), adjust `setup.sh` to not create `/etc/nginx/sites-available/sandstorm-php` going forward
- Make `/etc/nginx/sites-available/sandstorm-app` be a symlink to `/opt/sandstorm/.sandstorm/nginx.conf`

Backwards compatibility:

- I believe there's no backwards compatibility impact on existing vagrant-spk packages; they can enjoy their current situation.
- If they ever decide to refresh their `.sandstorm/*.sh` via `vagrant-spk setupvm lemp`, then they already expect to lose any customizations they've made to nginx.conf.
- For backwards compatibility clarity, in `setup.sh`, before we make a symlink to `/opt/app/.sandstorm/nginx.conf` we should check if the old filename (`/etc/nginx/sites-available/sandstorm-php`) exists on the filesystem. If it does, we should abort and tell the user they should either `rm` the old filename, or `vagrant destroy` then `vagrant up`.

zarvox:
`vagrant-spk setupvm <stack>` will now recursively copy that stack's `service-config` to `$PWD/.sandstorm/service-config`.

Then, we migrate the `static`, `lemp`, and `uwsgi` stacks to use this new functionality for their nginx confs, rather than doing messy patches atop Debian's default configs.

Tested successfully against:

* a minimal `static` stack "app" that just serves an `index.html`
* https://github.com/paulproteus/php-app-to-package-for-sandstorm
* https://github.com/zarvox/python-app-to-package-for-sandstorm

Closes #116.

### DIFF
--- a/stacks/lemp/service-config/mime.types
+++ b/stacks/lemp/service-config/mime.types
@@ -1,0 +1,89 @@
+
+types {
+    text/html                             html htm shtml;
+    text/css                              css;
+    text/xml                              xml;
+    image/gif                             gif;
+    image/jpeg                            jpeg jpg;
+    application/javascript                js;
+    application/atom+xml                  atom;
+    application/rss+xml                   rss;
+
+    text/mathml                           mml;
+    text/plain                            txt;
+    text/vnd.sun.j2me.app-descriptor      jad;
+    text/vnd.wap.wml                      wml;
+    text/x-component                      htc;
+
+    image/png                             png;
+    image/tiff                            tif tiff;
+    image/vnd.wap.wbmp                    wbmp;
+    image/x-icon                          ico;
+    image/x-jng                           jng;
+    image/x-ms-bmp                        bmp;
+    image/svg+xml                         svg svgz;
+    image/webp                            webp;
+
+    application/font-woff                 woff;
+    application/java-archive              jar war ear;
+    application/json                      json;
+    application/mac-binhex40              hqx;
+    application/msword                    doc;
+    application/pdf                       pdf;
+    application/postscript                ps eps ai;
+    application/rtf                       rtf;
+    application/vnd.apple.mpegurl         m3u8;
+    application/vnd.ms-excel              xls;
+    application/vnd.ms-fontobject         eot;
+    application/vnd.ms-powerpoint         ppt;
+    application/vnd.wap.wmlc              wmlc;
+    application/vnd.google-earth.kml+xml  kml;
+    application/vnd.google-earth.kmz      kmz;
+    application/x-7z-compressed           7z;
+    application/x-cocoa                   cco;
+    application/x-java-archive-diff       jardiff;
+    application/x-java-jnlp-file          jnlp;
+    application/x-makeself                run;
+    application/x-perl                    pl pm;
+    application/x-pilot                   prc pdb;
+    application/x-rar-compressed          rar;
+    application/x-redhat-package-manager  rpm;
+    application/x-sea                     sea;
+    application/x-shockwave-flash         swf;
+    application/x-stuffit                 sit;
+    application/x-tcl                     tcl tk;
+    application/x-x509-ca-cert            der pem crt;
+    application/x-xpinstall               xpi;
+    application/xhtml+xml                 xhtml;
+    application/xspf+xml                  xspf;
+    application/zip                       zip;
+
+    application/octet-stream              bin exe dll;
+    application/octet-stream              deb;
+    application/octet-stream              dmg;
+    application/octet-stream              iso img;
+    application/octet-stream              msi msp msm;
+
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document    docx;
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet          xlsx;
+    application/vnd.openxmlformats-officedocument.presentationml.presentation  pptx;
+
+    audio/midi                            mid midi kar;
+    audio/mpeg                            mp3;
+    audio/ogg                             ogg;
+    audio/x-m4a                           m4a;
+    audio/x-realaudio                     ra;
+
+    video/3gpp                            3gpp 3gp;
+    video/mp2t                            ts;
+    video/mp4                             mp4;
+    video/mpeg                            mpeg mpg;
+    video/quicktime                       mov;
+    video/webm                            webm;
+    video/x-flv                           flv;
+    video/x-m4v                           m4v;
+    video/x-mng                           mng;
+    video/x-ms-asf                        asx asf;
+    video/x-ms-wmv                        wmv;
+    video/x-msvideo                       avi;
+}
--- a/stacks/lemp/service-config/nginx.conf
+++ b/stacks/lemp/service-config/nginx.conf
@@ -1,0 +1,82 @@
+worker_processes 4;
+pid /var/run/nginx.pid;
+
+events {
+	worker_connections 768;
+	# multi_accept on;
+}
+
+http {
+	# Basic Settings
+	sendfile on;
+	tcp_nopush on;
+	tcp_nodelay on;
+	keepalive_timeout 65;
+	types_hash_max_size 2048;
+	# server_names_hash_bucket_size 64;
+	server_tokens off;
+	server_name_in_redirect off;
+
+	include mime.types;
+	default_type application/octet-stream;
+
+	# Logging
+	access_log off;
+	error_log stderr;
+
+	# Prevent nginx from adding compression; this interacts badly with Sandstorm
+	# WebSession due to https://github.com/sandstorm-io/sandstorm/issues/289
+	gzip off;
+
+	# Trust the sandstorm-http-bridge's X-Forwarded-Proto.
+	map $http_x_forwarded_proto $fe_https {
+		default "";
+		https on;
+	}
+
+	server {
+		listen 8000 default_server;
+		listen [::]:8000 default_server ipv6only=on;
+
+		# Allow arbitrarily large bodies - Sandstorm can handle them, and requests
+		# are authenticated already, so there's no reason for apps to add additional
+		# limits by default.
+		client_max_body_size 0;
+
+		server_name localhost;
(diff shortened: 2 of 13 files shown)


stacks/lemp/service-config/mime.types
```diff
@@ -0,0 +1,89 @@
+
```
paulproteus: I don't love having `mime.types` in `vagrant-spk`, but I guess it makes more sense than copying it in from the Debian VM.

stacks/lemp/service-config/nginx.conf
```diff
@@ -0,0 +1,82 @@
+worker_processes 4;
```
paulproteus: I do love having nginx in here.

It's semi-hilarious that we use `/var/run` when `/var/run` == `/var` for us, but I still am +1 on using `/var/run` since it looks more normal.

stacks/lemp/service-config/nginx.conf
```diff
@@ -0,0 +1,82 @@
...
+pid /var/run/nginx.pid;
+
+events {
+	worker_connections 768;
+	# multi_accept on;
+}
+
+http {
+	# Basic Settings
+	sendfile on;
```
paulproteus: I would probably disable `sendfile` but it's not a big deal.

stacks/lemp/service-config/nginx.conf
```diff
@@ -0,0 +1,82 @@
...
+	tcp_nopush on;
+	tcp_nodelay on;
+	keepalive_timeout 65;
+	types_hash_max_size 2048;
+	# server_names_hash_bucket_size 64;
+	server_tokens off;
+	server_name_in_redirect off;
```
paulproteus: If this config entry has meaning in Sandstorm, it's nice if we can add a comment. Not a show-stopper.

stacks/lemp/service-config/nginx.conf
zarvox: Why?  It can only improve performance in the cases where it's available, AFAIK.

paulproteus: (Not a show-stopper)

stacks/lemp/service-config/nginx.conf
zarvox: I'd rather use `/run` if `/run` were a tmpfs, the way it is guaranteed to be on modern distros, but it sounded like Kenton was unenthusiastic about having a second tmpfs mount in the sandbox.  So this is my compromise.

stacks/lemp/service-config/nginx.conf
```diff
@@ -0,0 +1,82 @@
...
+}
+
+http {
+	# Basic Settings
+	sendfile on;
+	tcp_nopush on;
+	tcp_nodelay on;
+	keepalive_timeout 65;
+	types_hash_max_size 2048;
+	# server_names_hash_bucket_size 64;
```
paulproteus: This commented-out config entry has been the source of many problems. Not a show-stopper, just curious if you understand http://nginx.org/en/docs/http/server_names.html yet. I don't yet.

stacks/lemp/service-config/nginx.conf
```diff
@@ -0,0 +1,82 @@
...
+	# Prevent nginx from adding compression; this interacts badly with Sandstorm
+	# WebSession due to https://github.com/sandstorm-io/sandstorm/issues/289
+	gzip off;
+
+	# Trust the sandstorm-http-bridge's X-Forwarded-Proto.
+	map $http_x_forwarded_proto $fe_https {
+		default "";
+		https on;
+	}
+
```
paulproteus: It sure would be nice if it were possible to share fractions of these config files between all stacks, but maybe `conf.d` etc is not worth it for app authors. So IMHO not a showstopper; I have no better solution; just wanted to remark on that.

stacks/lemp/setup.sh
```diff
@@ -5,37 +5,6 @@ set -euo pipefail
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get install -y nginx php5-fpm php5-mysql php5-cli php5-curl git php5-dev mysql-server
-rm -f /etc/nginx/sites-enabled/default
```
paulproteus: `/usr/bin/yes` to all this removal

stacks/lemp/service-config/mime.types
zarvox: Yeah.  The downside of vendoring in all of your configs is that you have to vendor in **all** of your configs, which may include a decent amount of stuff you don't really care about maintaining.

Here, I think the overhead is manageable, and it's also pretty easy to update the file from upstream.

stacks/lemp/setup.sh
```diff
@@ -5,37 +5,6 @@ set -euo pipefail
...
-    # WebSession due to https://github.com/sandstorm-io/sandstorm/issues/289
-    gzip off;
-
-    server_name localhost;
-    root /opt/app;
-    location / {
-        index index.php;
-        try_files \$uri \$uri/ =404;
-    }
-    location ~ \\.php\$ {
```
paulproteus: People frequently run into syntax error problems in the shell heredoc stuff, so I'm really excited we're getting rid of them!

stacks/static/service-config/mime.types
```diff
@@ -0,0 +1,89 @@
+
```
paulproteus: Given that this is 100% repeated from `stacks/lemp/service-config/mime.types` I wonder if we can use a symlink in the `stacks` directory, and then make sure `vagrant-spk` copies the file fully into the app.

paulproteus: I'm open to other ways to avoid duplication. Not necessarily a showstopper.

stacks/lemp/service-config/nginx.conf
paulproteus: See "Known Issues with sendfile(2)" on http://www.proftpd.org/docs/howto/Sendfile.html

Maybe I've spent too many years administering NFS and AFS.

paulproteus: Seems like in Sandstorm, we'll have no problem using sendfile, so sure, yay, OK! What a beautiful world we have.

vagrant-spk
```diff
@@ -511,6 +511,14 @@ def setup_vm(args):
     with open(vagrantfile_path, "w") as f:
         f.write(VAGRANTFILE_CONTENTS)
 
+    # Recursively copy in anything in the stack's service-config folder.
```
paulproteus: Since we're recursively copying, doing symlinks for `mime.types` should work OK. (But maybe not on Windows. But I can make the Windows installer do the copying, and we build the Windows installer on Linux anyway.)

stacks/lemp/service-config/nginx.conf
zarvox: I considered reusing files, but then that adds risk that changing one stack will break another, and it felt like the sort of thing where the upside of having an easy-to-read, straight-through config with clear invariants ("place files in folder, get files in folder") outweighed the downsides.

stacks/static/service-config/mime.types
zarvox: That would mean that either:

1. can't let the `service-config/` folder ship symlinks (which I think could be useful)
2. have to write some manifest format for apps to specify 

I mean, while throwing out ideas, we could also ship a symlink to `/etc/nginx/mime.types`, which is likely to have the desired effect, except that then we're relying on the Debian package layout again...

zarvox:
FWIW, I wrote this collection of configs to match the effective config produced by the various stacks, with the intent that this should not change behavior in a way that had the potential to break users.

If we want to optimize our nginx config settings, then I'm cool with doing so, but would prefer that happen in a separate PR.

vagrant-spk
zarvox: But then that's code complexity that we have to maintain.  I assert that the maintenance cost of this duplication is lower than the cost of implementing one of the alternatives.

paulproteus:
+1 to the discussion here. Thanks for your patience and :clap: for doing this.

Mostly I think we should document some of these design decisions somewhere, probably https://docs.sandstorm.io/en/latest/vagrant-spk/design/ .

zarvox:
Cheers.  Thanks for asking a bunch of reasonable, probing questions, to tease out the goals and tradeoffs made. :)

I guess this means I should also adapt the LESP stack that was recently contributed to this model too!

paulproteus:
I guess so! I can be the one to write docs & release notes when the time comes (hopefully/presumably this week).

PR #133, opened 2016-02-19
setup.sh: Provide a more intense warning & docs link

paulproteus:
Rationale:

- @grantpotter emailed me to ask why his vagrant-spk box wasn't being modified
  in accordance with his changes to setup.sh. I believe he didn't understand the
  comment text in the existing setup.sh because he doesn't have a clear concept
  of the boundary between vagrant and vagrant-spk.

- Therefore, I think he would have benefited from a more intense warning plus a
  link to the docs.

### DIFF
--- a/stacks/diy/setup.sh
+++ b/stacks/diy/setup.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
+
+# When you change this file, you must take manual action. Read this doc:
+# - https://docs.sandstorm.io/en/latest/vagrant-spk/customizing/#setupsh
+
 set -euo pipefail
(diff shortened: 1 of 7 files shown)


paulproteus:
Hi @zarvox - would you be willing to review, then +1/-1 this, and/or merge it?

Hi @grantpotter - do you think that if the top of `setup.sh` had said:

```bash
# When you change this file, you must take manual action. Read this doc:
# - https://docs.sandstorm.io/en/latest/vagrant-spk/customizing/#setupsh
```

you would have been able to solve your own problem?

If so, great. If not, then this pull request doesn't do what it needs to do, so I should figure out what _would_ have saved you the trouble!

PR #134, opened 2016-02-19
Add support for `spk keygen` and `spk getkey` commands

zarvox:
These are needed for the signing key migration instructions.


zarvox:
Approved by @paulproteus in over-the-shoulder review.

PR #137, opened 2016-02-21
Use PCNet-Fast III by default

### LINKED ISSUE #132: Improve networking speed
@synchrone noticed on IRC today that uploading to one's `vagrant-spk` base box can be very slow. In my manual testing this morning, on Debian, I reproduced the slowness and I also tried manually changing the network adapter type to `virtio` within the VirtualBox GUI. This massively improved network speed.

I'm going to try some experiments before suggesting a specific course of action.

paulproteus:
Close #132

Testing done:

- Asked @synchrone to manually verify on a Linux system: success.

- Asked @phildini to manually verify on a Mac system: success.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -114,6 +114,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     vb.cpus = cpus
     vb.memory = assign_ram_mb
     vb.name = VM_NAME
+    vb.customize ["modifyvm", :id, "--nictype1", "Am79C973"]
 
     override.vm.synced_folder "..", "/opt/app"
     override.vm.synced_folder ENV["HOME"] + "/.sandstorm", "/host-dot-sandstorm"


paulproteus:
@zarvox requesting your +1/-1/merge.

zarvox:
-0.

The test you used is quite possibly CPU-bound by SSH's stream cipher on either end (but particularly the guest), FWIW.

If we're going to go to the trouble of locking to older hardware, rather than letting VirtualBox/Vagrant be responsible stewards of that choice, I want to see numerical evidence that we're not shooting ourselves in the foot now and in the future.  I'd rather us leave the defaults unless we have affirmative evidence that this is better on average.

Further, I'd mildly prefer to use something that virtualizes gigabit hardware over the 10/100 PCNet-Fast III, since your test looks to be pretty close to the theoretical maximum transfer speed of the latter already, and having headroom there lets us make better use of hardware improvements in the future, even if the software remains the same.

I read the various other threads that you linked from #132, and I didn't see any numbers in there that led me to believe any particular network adapter outperforms any other under any circumstances.

I'm QUITE curious what the numbers looked like, since you didn't give any for VirtualBox on OSX (but you did give a bunch of numbers earlier in the thread that suggest the change makes no difference).

paulproteus:
Per IRL discussion between @zarvox and me, we have a consensus that since the difference here is night-and-day in manual testing, then it's worth merging the change even though we can't perceive it in the automated tests that I did.

paulproteus:
Merging this, which was (I believe) the outcome of the IRL conversation on Monday.

zarvox:
Yup, thanks.

PR #141, opened 2016-02-22
vagrant-spk: setupvm with no args should list supported stacks

### LINKED ISSUE #12: vagrant-spk should document a list of supported stacks
Right now it's not very discoverable what stacks exist and what they include.  If we want people to use these stacks, we should make them more visible.

zarvox:
Closes #12.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -477,6 +477,19 @@ def auto(args):
 
 
 def setup_vm(args):
+    if (len(args.command_specific_args) == 0):
+        # No stack specified.  List known stacks.
+        print("No stack specified.  Try specifying a stack with:")
+        print("")
+        print("vagrant-spk setupvm <stack>")
+        print("")
+        print("Supported stacks:")
+        stacks_dir = os.path.join(CODE_DIR, "stacks")
+        for stack in sorted(os.listdir(stacks_dir)):
+            print("  {}".format(stack))
+        print("")
+        return
+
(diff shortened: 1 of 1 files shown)


zarvox:
This behavior seems strictly superior to the previous behavior if you ran `vagrant-spk setupvm` with no additional arguments, which was to crash with an index out of range error.

vagrant-spk
```diff
@@ -477,6 +477,19 @@ def auto(args):
...
+        print("No stack specified.  Try specifying a stack with:")
+        print("")
+        print("vagrant-spk setupvm <stack>")
+        print("")
+        print("Supported stacks:")
+        stacks_dir = os.path.join(CODE_DIR, "stacks")
+        for stack in sorted(os.listdir(stacks_dir)):
+            print("  {}".format(stack))
+        print("")
+        return
```
paulproteus: In my opinon, this should `exit(1)` at the end. I'll merge as-is for latency-reduction reasons, but @zarvox if you're at least +0 on that, let me know and either you or I can make it `exit(1)`.

paulproteus:
+1 to this as-is; see comment for a non-showstopper suggestion.

paulproteus:
In fact, I can turn that suggestion into a pull request for you to merge/reject!

elimisteve:
Sweet! I agree that this is the right functionality/behavior.

PR #142, opened 2016-02-22
vagrant-spk: publish shouldn't crash if user gives no .spk arg

zarvox:
As observed by `synchrone` on IRC.


paulproteus:
+1

`example.spk` is a fine name to use here.

PR #145, opened 2016-02-26
Move all vagrant passthrough subcommands under the "vm" subcommand

zarvox:
In an attempt to clean up the command surface of vagrant-spk, all commands
intended to forward directly to vagrant are now placed under the "vm"
subcommand.

The "up", "ssh", "destroy", "global-status", and "reload" commands are
deprecated in favor of the "vm <subcommand>" equivalents, but will continue to
work with a warning until some future release.

This change also means we can reasonably pass through *any* arguments passed to
`vagrant-spk vm` on to `vagrant`, so we're no longer in the business of
whitelisting arguments and reimplementing all the commands that `vagrant`
provides that people might want to use.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -645,7 +645,15 @@ def publish(args):
         os.remove(temp_spk)
 
 def warn_deprecated(command_name):
-    sys.stderr.write("WARNING: `vagrant-spk {0}` is deprecated and may be removed in the future.\nPrefer `vagrant-spk vm {0}`.\n".format(command_name))
+    ANSI_RED = "\x1b[31m"
+    ANSI_RESET = "\x1b[0m"
+    message = "\n".join([
+      ANSI_RED + "WARNING: `vagrant-spk {0}` is deprecated and will be removed on August 1, 2016.".format(command_name),
+      "Prefer `vagrant-spk vm {0}`.".format(command_name),
+      ANSI_RESET + "For more information, see https://groups.google.com/forum/#!topic/sandstorm-dev/cuSNJ3IsP6I",
+      ""
+    ])
+    sys.stderr.write(message)
 
 def deprecated_halt(args):
     warn_deprecated("halt")
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -742,6 +742,9 @@ def main():
         Command('vm', vm_subcommand,
             "Pass through all subsequent arguments to `vagrant` run in `.sandstorm`."),
         Command('vm up', None, "Start the Vagrant VM (and build the VM image if needed)"),
+        Command('vm halt', None, "Shut down the Vagrant VM."),
+        Command('vm destroy', None, "Delete the Vagrant VM (but leave --work-directory untouched)."),
+        Command('vm ssh', None, "SSH into the Vagrant VM."),
         Command('init', init,
             "Initialize fresh Sandstorm package definition \n"
             "  resources in --work-directory."),
@@ -756,9 +759,6 @@ def main():
         Command('publish', publish,
             "Send the current packed app spk to \n"
             "  the Sandstorm App Market."),
-        Command('vm halt', None, "Shut down the Vagrant VM."),
-        Command('vm destroy', None, "Delete the Vagrant VM (but leave --work-directory untouched)."),
-        Command('vm ssh', None, "SSH into the Vagrant VM."),
         Command('keygen', keygen,
             "Generate a new key, add it to your keyring,\n"
             "  and print the identifier."),
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -565,6 +565,10 @@ def setup_vm(args):
     with open(stack_path, "w") as f:
         f.write(stack_plugin_name + "\n")
 
+def deprecated_bring_up_vm(args):
+    warn_deprecated("up")
+    bring_up_vm(args)
+
 def bring_up_vm(args):
     sandstorm_dir = os.path.join(args.work_directory, ".sandstorm")
     # Make sure ~/.sandstorm exists, since global-setup.sh uses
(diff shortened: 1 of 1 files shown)


vagrant-spk
```diff
@@ -640,22 +644,30 @@ def publish(args):
     finally:
         os.remove(temp_spk)
 
-def halt(args):
+def warn_deprecated(command_name):
+    sys.stderr.write("WARNING: `vagrant-spk {0}` is deprecated and may be removed in the future.\nPrefer `vagrant-spk vm {0}`.\n".format(command_name))
```
zarvox: Is this loud enough, or do we also want to use ANSI escape codes to make this red?

paulproteus: +1 to ANSI red, if you'd add it.

Also I suggest you say, "will be removed on August 1, 2016" rather than "may be removed [some day who knows whatever you can ignore this warning]".

If you prefer, I can be the one to make changes according to the above suggestions.

paulproteus: It'd also be nice to link to the email thread, or github issue, in this deprecation message, so people can easily access the context for the change.

https://groups.google.com/forum/#!topic/sandstorm-dev/cuSNJ3IsP6I is a link to the message in case that saves you the trouble of finding the link.

paulproteus:
Question about this: does `vagrant-spk vm up --provision` now successfully pass `--provision` through to `vagrant up`, or does one still need to `vagrant-spk vm up -- --provision`? I imagine it works OK, which is great if so.

I'm OK with me merging this before anyone updates the docs, though presumably that's important to do, too.

@zarvox let me know what you think of the above suggestions/requests. If you're at least +0 on them, I can try do them in follow-up commits, and if you're +1 on them and actually want to do them (or something like them), that'd be awesome.

vagrant-spk
```diff
@@ -676,6 +688,28 @@ def getkey(args):
...
+        sys.stderr.write("You can't call `vm` commands if there's no `.sandstorm` folder.\n" +
+                "Try `vagrant-spk setupvm <stack>` first.\n")
+        sys.exit(1)
+
+    subcommand = args.command_specific_args[0]
+    # Do subcommand specific sanity checks.
+    if subcommand == 'up':
+        # Sanity-check that the user will get a Vagrant box that supports vboxsf, since
+        # vagrant-spk depends on that for now.
+        ensure_working_vboxsf_in_base_box(sandstorm_dir)
```
paulproteus: Yay. When I add another sanity-check for `up`, namely that port 6080 is available (if I can detect that the Vagrantfile uses port 6080), I'll add that here.

vagrant-spk
```diff
@@ -676,6 +688,28 @@ def getkey(args):
...
             )
 
+def vm_subcommand(args):
+    sandstorm_dir = os.path.join(args.work_directory, ".sandstorm")
+    if len(args.command_specific_args) == 0:
+        sys.stderr.write("Try specifying a command to pass through to vagrant, like\n\nvagrant-spk vm up\n\n")
+        sys.exit(1)
+
+    # Do a couple sanity checks.
+    ensure_host_sandstorm_folder_exists()
```
paulproteus: Should `ensure_host_sandstorm_folder_exists()` do the printing+exiting that you do on L699?

vagrant-spk
```diff
@@ -705,24 +739,26 @@ def main():
...
+            "Pass through all following arguments to `vagrant`."),
+        ('halt', deprecated_halt,
+            "(deprecated, use 'vm halt') Shut down the Vagrant VM."),
+        ('destroy', deprecated_destroy,
+            "(deprecated, use 'vm destroy') Delete the Vagrant VM (but leave \n"
             "  --work-directory untouched)."),
-        ('global-status', global_status,
-            "Show the state of all active \n"
+        ('global-status', deprecated_global_status,
+            "(deprecated, use 'vm global-status') Show the state of all active \n"
```
paulproteus: suggest: s/vm global-status/vagrant-spk vm global-status/

Rationale: Otherwise, people might think "vm" is a command available on their $PATH.

Same suggestion elsewhere.

vagrant-spk
```diff
@@ -705,24 +739,26 @@ def main():
...
         ('keygen', keygen,
             "Generate a new key, add it to your keyring,\n"
             "  and print the identifier."),
         ('getkey', getkey,
             "Given a key identifier, retrieve and print the\n"
             "  corresponding private key from your keyring."),
-        ('reload', do_reload,
-            "Reboot the virtual machine, turning it off then on."),
+        ('reload', deprecated_do_reload,
+            "(deprecated, use 'vm reload') Reboot the virtual machine, turning it off then on."),
```
paulproteus: IMHO it's important that `vagrant-spk --help` print a list of likely-useful `vagrant-spk vm` commands. Curious if you agree.

zarvox: Hmm.  Maybe, but I'd really like it to be minimal.  The purpose of `vagrant-spk` (and its help options) is not to teach people how to vagrant.

Such a list should probably contain at most "up", "halt", and maybe "ssh".

vagrant-spk
zarvox: good point

vagrant-spk
zarvox: No.  `ensure_host_sandstorm_folder_exists()` checks for `~/.sandstorm`, not `<project_folder>/.sandstorm`.

If you think the check below is needed elsewhere, we can extract a `sanity_check_project_sandstorm_folder()` function or something like that.

zarvox:
I think this addresses all of your comments. :)

vagrant-spk
```diff
@@ -640,22 +644,38 @@ def publish(args):
     finally:
         os.remove(temp_spk)
 
-def halt(args):
+def warn_deprecated(command_name):
+    ANSI_RED = "\x1b[31m"
```
paulproteus: At some point we're going to have some difficulty porting this to Python 3. It's A-OK with me for now, though.

vagrant-spk
```diff
@@ -640,22 +644,38 @@ def publish(args):
...
+    ANSI_RESET = "\x1b[0m"
+    message = "\n".join([
+      ANSI_RED + "WARNING: `vagrant-spk {0}` is deprecated and will be removed on August 1, 2016.".format(command_name),
+      "Prefer `vagrant-spk vm {0}`.".format(command_name),
```
paulproteus: +1 to telling them what to do instead.

vagrant-spk
```diff
@@ -683,60 +734,69 @@ def main():
...
-        ('destroy', destroy,
-            "Delete the Vagrant VM (but leave \n"
-            "  --work-directory untouched)."),
-        ('global-status', global_status,
-            "Show the state of all active \n"
-            "  Vagrant VMs on this host."),
-        ('ssh', ssh,
-            "SSH into the Vagrant VM."),
-        ('keygen', keygen,
+        Command('vm halt', None, "Shut down the Vagrant VM."),
```
paulproteus: Exciting to see `None` here.

paulproteus:
+1, thanks so much for @zarvox for writing all this.

PR #147, opened 2016-03-16
Add instructions for using vagrant-spk with libvirt

zarvox:
This adds some instructions for how to set up `vagrant-spk` with the `libvirt` backend on Linux, rather than the VirtualBox one.  In particular, these instructions focus on Fedora, since that's what the author uses.

@dwrensha thanks for testing these out - let me know if you have any other changes/enhancements you'd suggest!


zarvox:
Self-merging because entirely documentation, and consensus was reached that publishing this set of instructions was a not-bad idea.

paulproteus:
+1 many thanks​

PR #148, opened 2016-03-21
libvirt HOWTO: Debian calls it libvirt-bin

paulproteus:
`libvirtd` is nonexistent, so I assume it's this. Will self-merge because we decided these docs need little review, but post-merge review very much welcome.


paulproteus:
Self-merging because 100% docs; cc: @zarvox for sanity-check

zarvox:
Actually, I think I was aiming for `libvirt-daemon-system`

paulproteus:
Fixed, thanks!

PR #151, opened 2016-03-31
Refine libvirt HOWTO for Debian

paulproteus:
Thanks to @zarvox for mentioning the right package.


paulproteus:
Self-merging because the libvirt HOWTO is basically a wiki, as I understand it.

PR #155, opened 2016-04-22
Fix executable path detection on Windows 10.

zarvox:
The pyinstaller bundle generates a program for which `sys.argv[0]` is not the
full path to the executable, but just the short name of the program.

In this case, we should use `sys.executable` to determine the path to the .exe
that is actually launching vagrant-spk, since that's where the stacks will be
located.

@paulproteus this fixes the issue hexx from IRC was having; if I can impose upon you to put together a Windows release, that'd be cool.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -32,7 +32,20 @@ import sys
 
 PWD = os.getcwd()
 ENABLE_EXPERIMENTAL = False
-CODE_DIR = os.path.dirname(os.path.realpath(sys.argv[0]))
+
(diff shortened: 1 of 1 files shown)


paulproteus:
Thanks @zarvox. I take it from your IRC conversations and this GitHub pull request that you've manually tested this fix on Windows. Thanks for doing that!

I'll see now about doing a new release.

paulproteus:
In general, feel free to self-merge these things if you think it's important; but also I'm always happy to review and click buttons.

Issue #158, opened 2016-04-27
Can we please do something about this error message?


ndarilek:
There is a certain vagrant-spk failure case that happens sometimes. I'm including the log here in its entirety. I don't know what it says because, for some reason, the error message is split up one character per line. Maybe visually it's a bit odd but you can scan the text in a column, but with a screen reader I have to listen to the entire line, hear the single character, and string them together in my head. I don't have the patience to do that. :) I've seen this error in a few vagrant-spk-generated stacks. This specimen is from the Sandstorm Radicale app at https://github.com/synchrone/sandstorm-radicale. Literally just cloned and ran `vagrant-spk up`.

```
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: stdin: is not a tty
==> default: Downloading Sandstorm version 160...
==> default: c
==> default: u
==> default: r
==> default: l
==> default: :
==> default:  
==> default: (
==> default: 1
==> default: 8
==> default: )
==> default:  
==> default: t
==> default: r
==> default: a
==> default: n
==> default: s
==> default: f
==> default: e
==> default: r
==> default:  
==> default: c
==> default: l
==> default: o
==> default: s
==> default: e
==> default: d
==> default:  
==> default: w
==> default: i
==> default: t
==> default: h
==> default:  
==> default: 2
==> default: 6
==> default: 4
==> default: 0
==> default: 4
==> default: 2
==> default: 7
==> default:  
==> default: b
==> default: y
==> default: t
==> default: e
==> default: s
==> default:  
==> default: r
==> default: e
==> default: m
==> default: a
==> default: i
==> default: n
==> default: i
==> default: n
==> default: g to read
```

Seems to happen when attempting to download Sandstorm? I have no idea what causes it because I just don't have time to piece together the characters, but it's odd how Vagrant seems to mangle the output.

Thanks.

paulproteus:
Yes, definitely. Thanks for filing the bug. For what it's worth, it seems
to say:

curl: (18) connection closed with 2640427 bytes remaining to read.

So it seems like some kind of slow/semi-broken networking issue. We should
probably allow curl more time to do the download. Having said that, I am
curious - if you adjust the Vagrantfile to use NIC type Am79C973 (see
https://github.com/sandstorm-io/vagrant-spk/blob/master/vagrant-spk#L125 )
does that help?

Anyway I can work on fixing the scripts for this. Sadly this is part of
[...]

PR #159, opened 2016-04-29
For some calls to curl, buffer output via pipe to cat

paulproteus:
Close: #158 

## Explanation of this implementation strategy

This changeset:

- Adjusts the default `global-setup.sh` so that the two `curl` calls that currently only print error messages now print them error messages through a pipe to cat.

- This has the effect of buffering them, which helps vagrant receive the error messages all at once, in my manual testing.

- Assuming this is implemented correctly, this retains all the same information as before. Any error messages that we were printing before, we are still printing. I know that @zarvox is interested in making sure that we retain the existing discoverability and comprehensibility of `vagrant-spk`.

## Other research

Here is my proof that the very subtle change on the first line is a safe change to make.

```
paulproteus@slittingmill:~$ cat << EOF > /tmp/print-to-both.sh
> #!/bin/bash
> echo stdout
> echo stderr 1>&2
> EOF
paulproteus@slittingmill:~$ bash /tmp/print-to-both.sh 
stdout
stderr
paulproteus@slittingmill:~$ bash /tmp/print-to-both.sh  > /dev/null
stderr
paulproteus@slittingmill:~$ bash /tmp/print-to-both.sh 2>&1 > /dev/null | cat
stderr
paulproteus@slittingmill:~$ bash /tmp/print-to-both.sh 2>&1 > /tmp/some-filename | cat
stderr
paulproteus@slittingmill:~$ cat /tmp/some-filename 
stdout
paulproteus@slittingmill:~$ 
```

### To reproduce the problem

Modify `CURL_OPTS` to also add ` --max-time 1` for a maximum wait of 1 second for the download.

## Next steps

@zarvox if you're willing to be the one to review/merge this, I'd appreciate that. Otherwise I might self-merge on Friday evening.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -162,13 +162,18 @@ GITIGNORE_CONTENTS = r"""
 GLOBAL_SETUP_SCRIPT = r"""#!/bin/bash
 set -euo pipefail
 
-# Set options for curl. Since we only want to show errors from these curl
-# commands, we also use 'cat' to buffer the output; for more information:
+# Set options for curl. Since we only want to show errors from these curl commands, we also use
+# 'cat' to buffer the output; for more information:
 # https://github.com/sandstorm-io/vagrant-spk/issues/158
+
 CURL_OPTS="--silent --show-error"
 echo localhost > /etc/hostname
 hostname localhost
+
+# The following line copies stderr through stderr to cat without accidentally leaving it in the
+# output file. See: https://github.com/sandstorm-io/vagrant-spk/pull/159
 curl $CURL_OPTS https://install.sandstorm.io/ 2>&1 > /host-dot-sandstorm/caches/install.sh | cat
+
 SANDSTORM_CURRENT_VERSION=$(curl $CURL_OPTS -f "https://install.sandstorm.io/dev?from=0&type=install")
 SANDSTORM_PACKAGE="sandstorm-$SANDSTORM_CURRENT_VERSION.tar.xz"
 if [[ ! -f /host-dot-sandstorm/caches/$SANDSTORM_PACKAGE ]] ; then
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -162,15 +162,18 @@ GITIGNORE_CONTENTS = r"""
 GLOBAL_SETUP_SCRIPT = r"""#!/bin/bash
 set -euo pipefail
 
+# Set options for curl. Since we only want to show errors from these curl
+# commands, we also use 'cat' to buffer the output; for more information:
+# https://github.com/sandstorm-io/vagrant-spk/issues/158
 CURL_OPTS="--silent --show-error"
 echo localhost > /etc/hostname
 hostname localhost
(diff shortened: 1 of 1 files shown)


paulproteus:
UPDATE: To reproduce the problem, you must also do something like delete the cached Sandstorm `tar.xz` in (I believe) `~/.sandstorm/caches/`.

paulproteus:
For @dwrensha - can you reproduce my claim that these changes make your output less verbose?

For @ndarilek - This was the best I could do to fix the problem you encountered. Assuming that it works, then in the case of curl having error messages that result in long multi-line output in the future, you can apply this patch to their `global-setup.sh`: https://github.com/sandstorm-io/vagrant-spk/commit/31c223ea1b897631221eaed9c4d311807e7adfc3.patch e.g. `curl https://github.com/sandstorm-io/vagrant-spk/commit/31c223ea1b897631221eaed9c4d311807e7adfc3.patch | git am`

vagrant-spk
```diff
@@ -162,15 +162,18 @@ GITIGNORE_CONTENTS = r"""
...
 set -euo pipefail
 
+# Set options for curl. Since we only want to show errors from these curl
+# commands, we also use 'cat' to buffer the output; for more information:
+# https://github.com/sandstorm-io/vagrant-spk/issues/158
 CURL_OPTS="--silent --show-error"
 echo localhost > /etc/hostname
 hostname localhost
-curl $CURL_OPTS https://install.sandstorm.io/ > /host-dot-sandstorm/caches/install.sh
+curl $CURL_OPTS https://install.sandstorm.io/ 2>&1 > /host-dot-sandstorm/caches/install.sh | cat
```
zarvox: I'm a little nervous about this - what happens in, say, the 404 case when you send `2>&1` and redirect stdout to a file and also pipe that to `cat`?  Is there a way to express this in a way that looks less like you're trying to redirect the same stdout to two places at once?

paulproteus: My test bash heredoc in the pull request demonstrates that this does what I think it does, fwiw.

If you prefer, I can set up some named pipes with `mkfifo` in the script. I experimented with this for a while, and sadly I can't find any other similarly-short invocation that also works.

Let me know what you think.

zarvox: Oh, I see that's the proof exists to show that this is in fact okay.

I guess I'd appreciate a comment on the fact that it's a very tricksy line.

paulproteus: +1 on comment that it's tricksy​

zarvox: Hah, timing collision.

Leave the code as-is (I don't really want to drag named pipes into this), but add a comment discouraging changing that line without fully understanding it.

Then, merge.

paulproteus:
Thanks @zarvox ! I went maybe a little overboard with commenting, but I think that's OK in this case. I'll merge. Thanks for your review; I appreciate it.

dwrensha:
This does seem to make my output less verbose!

paulproteus:
Wa-hoo!​

PR #160, opened 2016-04-30
fix deprecation warning

astraw:
This corrects the deprecation warning so that the suggested command runs without failure, at least on Vagrant 1.8.1


zarvox:
Thanks!

PR #163, opened 2016-05-07
Update example usage

pgrm:
`vagrant-spk up` is deprecated `vagrant-spk vm up` should be used instead


zarvox:
Thanks for the fix!

Issue #56, opened 2015-08-08
Provide a way to "enter" a running grain


paulproteus:
As an app developer, it would be nice to be able to get a shell inside a running grain to figure out why something is acting the way it's acting.

The standard recommendation is to run:

```
vagrant-spk ssh
cd /opt/app/.sandstorm
sudo nsenter --target $(pidof sandstorm-http-bridge) --wd --mount --net --ipc --uts --pid
```

This issue tracks the creation of a `vagrant-spk` command (e.g. `vagrant-spk entergrain {{grainid}}`) to achieve that.

Separately, it would be nice if doing so would not grow the `sandstorm-files.list`. Since (I believe) that requires cooperation from `spk`, I also filed https://github.com/sandstorm-io/sandstorm/issues/669

JamborJan:
What is the status on that now or in other words: what do you expect from the community? I have been using two things @zarvox told me:

First connecting to a mysql database inside a grain like this:
```
vagrant-spk ssh
sudo -i
mysql -u root --socket
/opt/sandstorm/var/sandstorm/grains/$YOUR_GRAIN_ID/sandbox/run/mysqld/mysqld.sock
```
As I have been working on mysql based apps this was super useful.

And also:
```
cd git/vagrant-spk/
git pull
git checkout nsenter-devgrain
cd git/your-work-in-progress-package
# make sure sandstorm in that VM is running a single grain
vagrant-spk devjoin
```
I only used this once or so as I rather needed access to the database to check things. 

I guess you are planning to build a wrap around that which is valid for everything: access logs in the grain, data and so on.

There is one question I have to clarify for me: what about things like tools for debugging. For example you don't necessarily need a mysql-cli installed to have a running grain e.g. when php-mysql does everything. But to access the database with an client application via ```vagrant-spk entergrain {{grainid}}``` I could use my local installed mysql client no matter if I have a Windows, Mac OS X or Linux machine. Did I get this right?

ndarilek:
Is there any way to enter the grain as the Sandstorm UID/GID? I tried passing `-U 998 -G 998` to the nsenter command line above but am always dropped in as root. 998 is the UID/GID of sandstorm on my VM. I'm trying to diagnose some permissions problems, so being able to pop in as root, change a permission, pop in as sandstorm and run whatever failed would be helpful.

kentonv:
[...]

PR #165, opened 2016-05-11
Add v1 of vagrant-spk devjoin

paulproteus:
Implementation notes:

- This adds an executable binary to the git repo, and its
  rust source code. It needs a rust nightly compiler at the moment.

- It injects that binary into /tmp into the namespace of the grain, then
  executes it.

- This version does not actually need to live within /tmp of the grain, but I need
  something similar in the near future for when I am injecting busybox.

- This is only available if you set VAGRANT_SPK_EXPERIMENTAL=Y in your environment
  for now.

- User-facing documentation is forthcoming.

Huge thanks to @zarvox for discussing possible designs.


paulproteus:
Hi @zarvox,

I'm hoping you'll be willing to click merge on this, and also tell me if there are important things to do next.

I think it's important to:

- Check for bash and 'ls' within the grain's filesystem, and if bash and/or ls are missing, exec busybox instead

- Check that it works properly on Windows. (e.g. does this helpers path actually get bundled into the installer? Do I accidentally corrupt the `enter_grain` binary due to corrupting it?

- Decide if `devjoin` is the right name for this.

- Publish a PR for docs.

- Handle your feedback on the embedded `sh` snippets here, or on the Python code I wrote.

Arguably-important but not urgent:

- Clean up my rust source so it's less silly-looking.

I hope you'll take a moment to manually test this, but you don't have to, but it'd be cool if you did since I'm proud that this monstrosity works at all.

vagrant-spk
```diff
@@ -620,6 +673,125 @@ def dev(args):
...
+    enter_grain_binary_path = tmp_vagrant_spk_dir + '/enter-grain'
+    commands = ' && '.join([
+        # Make sure /tmp/vagrant-spk exists in the VM.
+        'sudo mkdir -p {}'.format(tmp_vagrant_spk_dir),
+        # If the sha1sum tool exists, and the binary in this directory has the right sha1sum,
+        # then skip sending it over.
+        'if which sha1sum >/dev/null && sudo sha1sum {} 2>/dev/null | grep -q ^{} ; then sudo chmod 755 {} && exit 0 ; fi'.format(
+            enter_grain_binary_path, enter_grain_checksum, enter_grain_binary_path),
+        # If not, then use cat to inject it in.
+        'sudo dd of={} 2>/dev/null'.format(enter_grain_binary_path),
```
zarvox: This doesn't quite match the comment - is there a reason to prefer one or the other?

paulproteus: I'll fix the comment.

I use `dd` not `cat` because `dd` can open a file itself, and I need the file opened as root, so `sudo cat > filename` will give EPERM.

zarvox: Ahh, sounds good.

Random: a pattern that I use when working interactively to solve the same problem (write this file as root) is `cat | sudo tee filename`, with the side-effect of duplicating the output to stdout (which you probably don't want here).

paulproteus: Yeah, I never liked that aspect of `sudo tee`, alas. :)

paulproteus: i,i also unnecessary use of cat award, for you :trophy:

paulproteus: (I attempted to deliver that award lovingly, not meanly, so apologies if it comes off as mean.)

paulproteus:
Note for review:

- I've renamed this to `vagrant-spk shell` per feedback from Kenton.

paulproteus:
I've also un-marked it as being EXPERIMENTAL. It seems to work OK.

paulproteus:
Refs #56

paulproteus:
Per discussion with @zarvox I am:

- marking this as requiring `VAGRANT_SPK_EXPERIMENTAL=Y` in the environment, and

- self-merging now.

helpers/enter-grain-source/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+        syscall!(CLOSE, fd);
+    }
+}
+
+fn sanity_check_fd(fd: usize) {
+    // For some reason, syscall!(OPEN) returns usize, but I need to check its output against
+    // -1. So I'm going to just check if it's >255.
+    if fd > 255 {
+        let exit_failure = 1usize;
+        write(1, ("Failed to open a needed file. Bailing.\n".as_bytes()));
```
zarvox: More idiomatic would be to `panic!` here.

helpers/enter-grain-source/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+    setns(uts_fd, 0x04000000usize); // CLONE_NEWUTS
+    close(uts_fd);
+    setns(net_fd, 0x40000000usize); // CLONE_NEWNET
+    close(net_fd);
+    setns(pid_fd, 0x20000000usize); // CLONE_NEWPID
+    close(pid_fd);
+    setns(mnt_fd, 0x00020000usize); // CLONE_NEWNS which I guess is mount namespaces
+    close(mnt_fd);
+    fchdir(cwd_fd);
+    close(9);
```
zarvox: This should probably be `cwd_fd`, not `9`.

helpers/enter-grain-source/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+    // fork, and do a handful of things in the child before we execve bash.
+    let fork_result = fork();
+    if fork_result == 0 {
+        // in the child
+        setuid_setgid_1000();
+        write(1, "About to start shell...\n\0".as_bytes());
```
zarvox: I might remove this line.

paulproteus:
Other than these comments, is this OK to to switch to non-experimental, in your view? Earnestly asking. I'll work on these changes this week; I'm also aiming to know how much work it will take me to get this ready for non-experimental status so that I can plan accordingly.

You'll probably also want to review my docs changes; those are in the Sandstorm repo: https://github.com/sandstorm-io/sandstorm/pull/1979

My guess is that you'll have stronger opinions about the docs than the code, which I think is A-OK; I want to make sure to allocate enough time to be responsive to your feedback, so the sooner you can review that, the better. I'm also OK with that review process being IRL with me taking notes, in case that makes things easier for you, or whatever other process is easier for you and works for me.

helpers/enter-grain-source/src/main.rs
paulproteus: +1, will resolve that in upcoming PR

helpers/enter-grain-source/src/main.rs
paulproteus: +1, will resolve that in upcoming PR

PR #169, opened 2016-05-26
Add helpers to Windows build

paulproteus:
Also pin to version 3.1.1 of pyinstaller, which manages to build our program OK.


paulproteus:
Self-merging because 100% windows-build stuff.

PR #170, opened 2016-05-27
Add CHANGELOG for v0.164

paulproteus:


paulproteus:
Self-merging because 100% changelog.

PR #173, opened 2016-06-03
Fix enter-grain, which was broken in all builds

paulproteus:
There were a few problems. They're detailed in these commits. I'm going to self-merge and do a release, since the status quo is broken.


paulproteus:
@zarvox when you read this, you're going to love it.

PR #178, opened 2016-06-10
Reduce the size of `enter_grain` by 80%

zarvox:
With this change, we:

* enable link-time optimization
* use the system allocator over jemalloc
* strip the binary to remove debugging symbols

This reduces the binary size from 644k to 131k.

@paulproteus I didn't update the checked-in `enter_grain` binary or shasum; let me know if you'd like me to do so.  Tragically, any changes we make just make the git tree larger, so I'm not sure checking in non-bugfixes for the binaries is actually a net win.

### DIFF
--- a/helpers/Makefile
+++ b/helpers/Makefile
@@ -3,7 +3,7 @@ all: enter_grain.sha1
 .PHONY: enter_grain
 
 enter_grain:
-	cd enter-grain-source && cargo build --release && cp target/release/enter_grain ../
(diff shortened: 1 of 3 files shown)


paulproteus:
I think we're not going to change this binary all that much (famous last
words, I know), so I propose that you manually test your new binary, and if
it passes, you update the *.sha1 and the enter_grain binary in the repo,
and you self-merge.

++ for your minification work.
​

paulproteus:
Merging; when we next update the code, we can ship the newly-minified version. Thanks @zarvox !

PR #187, opened 2016-09-08
Move MySQL tmpdir to /var/tmp so it doesn't hit the 16MB tmpfs limit

zarvox:


stacks/lemp/setup.sh
```diff
@@ -30,9 +30,10 @@ sed --in-place='' \
 sed --in-place='' \
         --expression='s/^;clear_env = no/clear_env=no/' \
         /etc/php5/fpm/pool.d/www.conf
-# patch mysql conf to not change uid
+# patch mysql conf to not change uid, and to use /var/tmp over /tmp
 sed --in-place='' \
         --expression='s/^user\t\t= mysql/#user\t\t= mysql/' \
+        --expression='s/^tmpdir\t\t= \/tmp/tmpdir\t\t= \/var\/tmp/' \
```
paulproteus: IMHO this is very hard to read with `/` as the delimeter. Would you possibly be willing to use `s,foo,bar,` syntax i.e. use `,` or another non-`/` character as the delimeter?

zarvox: Done

paulproteus:
:clap: :tada:

paulproteus:
Thanks! I'm grateful. I've made a note to do a vagrant-spk release this Friday.

PR #190, opened 2016-09-16
Write CHANGELOG for v0.185

paulproteus:


paulproteus:
Self-merging because 100% CHANGELOG. @zarvox and others, feedback welcome.

PR #191, opened 2016-09-21
This will be v0.186 instead

paulproteus:


paulproteus:
Waiting to do a release means I have to change the version number. :)

Issue #195, opened 2016-11-17
Packaging example don't work as expected, because mysql server won't start


swarnat:
Hy,
I'm not sure, this is the correct place, but I don't know a better one. I also try to find already reported problems, but don't found similar topics.

I try to start to play with packaging the example app ([php-app-to-package-for-sandstorm](https://github.com/sandstorm-io/php-app-to-package-for-sandstorm))  a little bit and run every command from Documentation. Everything works well, except the App itself.
Sandstorm only shows the loading spinner and nothing was shown.

Because I would get it running, I start to debug (I must say Vagrant is a great software. Never know this before in this detail)

The log of the DEV grain looks like:
```
** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns
Installing MySQL system tables...
161117 18:21:25 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.
161117 18:21:25 [Note] Ignoring --secure-file-priv value as server is running with --bootstrap.
161117 18:21:25 [Note] /usr/sbin/mysqld (mysqld 5.5.53-0+deb8u1) starting as process 29 ...
OK
Filling help tables...
161117 18:21:25 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.
161117 18:21:25 [Note] Ignoring --secure-file-priv value as server is running with --bootstrap.
161117 18:21:25 [Note] /usr/sbin/mysqld (mysqld 5.5.53-0+deb8u1) starting as process 35 ...
OK

To start mysqld at boot time you have to copy
support-files/mysql.server to the right place for your system

....remove not relevant lines...

cd /usr/mysql-test ; perl mysql-test-run.pl

Please report any problems at http://bugs.mysql.com/

waiting for mysql to be available at /var/run/mysqld/mysqld.sock
161117 18:21:26 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.
/usr/sbin/mysqld: Error on realpath() on '/var/lib/mysql-files' (Error 2)
161117 18:21:26 [ERROR] Failed to access directory for --secure-file-priv. Please make sure that directory exists and is accessible by MySQL Server. Supplied value : /var/lib/mysql-files
161117 18:21:26 [ERROR] Aborting

[...]

PR #196, opened 2016-11-24
Add secure-file-priv to my.conf

JamborJan:
This change adds secure-file-priv to my.conf during setup.sh. For details see: https://github.com/sandstorm-io/vagrant-spk/issues/195


dwrensha:
Works for me!

PR #199, opened 2017-02-27
Switching to node.js v6 LTS

JamborJan:
Switching to node.js v6 LTS. This will solve a lot of dependency issues for most users which want to run the node stack out of the box.


zarvox:
Subsumed by #198

PR #200, opened 2017-03-17
Catch and quietly discard useless stacktrace

ndarilek:


zenhack:
@ndarilek, as mentioned on the mailing list, we recently merged #203, which fixes the same issue -- sorry again for not noticing that this pr was open. Is that patch a suitable replacement for this one in your opinion? If so we should close this.

//cc @zarvox, @paulproteus

ndarilek:
Sure, that works. Thanks, closing.

PR #201, opened 2017-03-27
Update Vagrant-SPK to support Windows10 hosts

JonTheNiceGuy:
On my Windows10 machine, running vagrant-spk vm up results in a stack trace. This is because I can't bind to 0.0.0.0.

Instead, forcably set it to bind to 127.0.0.1 - all the functionality, none of the issues (for me, at least!)

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -74,7 +74,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # We forward port 6080, the Sandstorm web port, so that developers can
   # visit their sandstorm app from their browser as local.sandstorm.io:6080
   # (aka 127.0.0.1:6080).
-  config.vm.network :forwarded_port, guest: 6080, host: 6080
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Interesting. I don't recall having to do anything like this when I tested vagrant-spk and I was definitely on Windows 10 at the time.

JonTheNiceGuy:
I got a big blob of python stack trace, and right in the middle was something about "can't bind 0.0.0.0". From this I determined that this binding to ports was the issue.

```
C:\Users\Jon Spriggs>"\Program Files\Oracle\VirtualBox\VBoxManage.exe" --version
5.1.18r114002

C:\Users\Jon Spriggs>vagrant --version
Vagrant 1.9.3
```

That said, if I'm not right, what does binding to 127.0.0.1 matter? :)

ocdtrekkie:
Yeah, as far as I can tell, there's no reason *not* to specify 127.0.0.1 in here. I just found it interesting because I had definitely tested it on Windows 10.

PR #203, opened 2017-04-21
Better error messages when a command fails.

zenhack:
Typically the command itself prints something reasonable, but we weren't
catching the exception, so this was obscured by a useless python stack
trace. We now catch the exception and print a concise message.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -291,7 +291,13 @@ def call_vagrant_command(sandstorm_dir, *command_args):
     command = ["vagrant"]
     command.extend(command_args)
     sys.stderr.write("Calling {} in {}\n".format(" ".join(["'{}'".format(arg) for arg in command]), sandstorm_dir))
-    return subprocess.check_call(command, cwd=sandstorm_dir)
+    try:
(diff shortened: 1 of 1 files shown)


vagrant-spk
```diff
@@ -291,7 +291,13 @@ def call_vagrant_command(sandstorm_dir, *command_args):
     command = ["vagrant"]
     command.extend(command_args)
     sys.stderr.write("Calling {} in {}\n".format(" ".join(["'{}'".format(arg) for arg in command]), sandstorm_dir))
-    return subprocess.check_call(command, cwd=sandstorm_dir)
+    try:
+        return subprocess.check_call(command, cwd=sandstorm_dir)
+    except subprocess.CalledProcessError as e:
+        ANSI_RED = "\x1b[31m"
+        ANSI_RESET = "\x1b[0m"
+        msg = "Command failed with a non-zero exit status (%d)." % e.returncode
```
zarvox: nit: generally I prefer the [new-style format strings](https://docs.python.org/2/library/string.html#format-string-syntax) over printf-style `%` formatting

zarvox:
Thanks! :)

PR #206, opened 2017-04-26
Don't include the shell in the spk for go stacks.

zenhack:


zarvox:
In hindsight, you can accomplish this (less conveniently, perhaps) by editing the `sandstorm-pkgdef.capnp` after calling `spk init` to directly launch `/opt/app/app` rather than `/opt/app/.sandstorm/launcher.sh`.

@zenhack Do you mind if I revert this and merge #184 so things can work out-of-the-box for Windows users?  As noted above, you can still achieve the "avoid including bash in the package" outcome.

PR #209, opened 2018-02-20
temp fix for broken vagrant box link

dmitry-sher:
added a solution as described here:
https://github.com/hashicorp/vagrant/issues/9442

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -58,9 +58,13 @@ VM_NAME = File.basename(File.dirname(File.dirname(__FILE__))) + "_sandstorm_#{Ti
 # Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
 
+# ugly hack to prevent hashicorp's bitrot. See https://github.com/hashicorp/vagrant/issues/9442
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
+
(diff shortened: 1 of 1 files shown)


dmitry-sher:
kudos to @kasgel and @kentonv , the PR is now much more mature

Rivierg:
Hi, I’m trying to install sandstorm using vagrant-spk for the first time and your fix really helped me, thanks a lot !
One thing before merging though, it is breaking the compatibility with the LEMP stack. In [https://github.com/sandstorm-io/vagrant-spk/blob/master/stacks/lemp/setup.sh](https://github.com/sandstorm-io/vagrant-spk/blob/master/stacks/lemp/setup.sh), php5 is to be installed. With the Debian version you specify in this file, it should be php7.0 ... I am trying to check it now see if I can fix it, then I will upload it here.

dmitry-sher:
@Rivierg do you want me to include #211 in #209 ?
@kentonv can somebody at some point really merge PRs?

ocdtrekkie:
Kenton will probably check in again on the weekend.

I think the best case is likely to keep 209 and 211 separate so people get credit for their contribution, and they arguably fix different issues, one being the actual operation of vagrant-spk and one fixing a stack.

Rivierg:
I agree that it should probably stay separated. As @ocdtrekkie said in #211 , handling the different OS and php versions could be done by the user (but it’s not great for a first tutorial, so we rather need to refine the documents for php).

kentonv:
Thanks!

PR #210, opened 2018-02-24
Updated the box to debian/contrib-jessie64

kasgel:
The sandstorm/debian-jessie64 VM Box is no longer available for download.
Switched to an alternative box debian/contrib-jessie64.

The new box does not come with curl by default, so the installation line of curl was also added.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -60,7 +60,7 @@ VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Base on the Sandstorm snapshots of the official Debian 8 (jessie) box.
-  config.vm.box = "sandstorm/debian-jessie64"
+  config.vm.box = "debian/contrib-jessie64"
 
   if Vagrant.has_plugin?("vagrant-vbguest") then
     # vagrant-vbguest is a Vagrant plugin that upgrades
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
This should be preferred over #209 since we figured out that doesn't actually solve the problem since the box cannot be downloaded. Note to visiting guests: Building with vagrant-spk is still broken on https://github.com/capnproto/capnproto/issues/635

kentonv:
I wonder if we should upgrade from jessie to stretch while we're at it?

ocdtrekkie:
IMHO we should make sure we get people able to build apps again before we risk breaking more things, but I like updating things. (Someday I'll get a machine built again to test this directly, mind you. I'd need a release/Windows build to test myself otherwise.)

kentonv:
In my experience, switching to stretch "just works". Can we try it? Otherwise I fear no one will ever update it later...

ocdtrekkie:
@kasgel @dmitry-sher @xet7 Do any of you want to try out debian/contrib-stretch64?

dmitry-sher:
Checking it now.
As far as i can see, you need both #209 and #210 combined to run `debian/contrib-jessie64`. without the url fix from #209 download fails for fresh systems without downloaded vbox.
I will check the `debian/contrib-stretch64` as soon as i get a working copy of jessie

dmitry-sher:
I confirm that #209 and #210 together with today's fixes from @kentonv fix everything: vm creation, dev mode, and packaging.

Will test now stretch.

dmitry-sher:
I confirm that stretch with current version=9.3.0 is also working. At least for our python apps. Have not yet tested meteor ones, but they don't use vagrant-spk at all, so it should not be a problem. 

I've updated #209 to include everything: url fix, box image fix.

PR #211, opened 2018-02-28
Update php version in LEMP stack to match updated debian version.

Rivierg:
Hi, this is a suggested fix to complement #209 , which updated the Debian version, breaking the php version dependency in the LEMP stack.
I tested it locally and it works for me when installing Sandstorm. However this has potential huge consequences for users requiring php5, so I would suggest that main developers check this before merging (I don’t have any experience with Sandstorm or vagrant yet).

### DIFF
--- a/stacks/lemp/setup.sh
+++ b/stacks/lemp/setup.sh
@@ -7,29 +7,29 @@ set -euo pipefail
 
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
-apt-get install -y nginx php5-fpm php5-mysql php5-cli php5-curl git php5-dev mysql-server
+apt-get install -y nginx php7.0-fpm php7.0-mysql php7.0-cli php7.0-curl git php7.0-dev mysql-server
 service nginx stop
-service php5-fpm stop
+service php7.0-fpm stop
 service mysql stop
 systemctl disable nginx
-systemctl disable php5-fpm
+systemctl disable php7.0-fpm
 systemctl disable mysql
-# patch /etc/php5/fpm/pool.d/www.conf to not change uid/gid to www-data
+# patch /etc/php/7.0/fpm/pool.d/www.conf to not change uid/gid to www-data
 sed --in-place='' \
         --expression='s/^listen.owner = www-data/;listen.owner = www-data/' \
         --expression='s/^listen.group = www-data/;listen.group = www-data/' \
         --expression='s/^user = www-data/;user = www-data/' \
         --expression='s/^group = www-data/;group = www-data/' \
-        /etc/php5/fpm/pool.d/www.conf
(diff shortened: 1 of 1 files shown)


Rivierg:
I also noticed, this change would imply to update the doc in https://docs.sandstorm.io/en/latest/vagrant-spk/customizing/#setupsh.

ocdtrekkie:
Assuming this gets merged at some future point, the docs file to edit would be here: https://github.com/sandstorm-io/sandstorm/edit/master/docs/vagrant-spk/customizing.md

My general opinion is that app updates made with vagrant-spk right now are extremely rare, due to ordinary bitrot, anyone wishing to upgrade a PHP app will already have to dodge significant pain to build again. I know @xet7 is interested in messing with WordPress, which recommends PHP 7.2 as it is.

I'd prefer if it was possible for vagrant-spk to support both php5 and php7 at the same time (many web hosts allow you to configure it by folder), but if the Debian version is going to force our hand, that may just be the right way forward. (As it is, nothing is preventing someone from changing their vagrant-spk to a different Linux version, different PHP version, etc., if they really need an old PHP version, and the Docs could always note this as a footnote somewhere.)

Rivierg:
OK, I see, it would be better indeed to change the vagrant file and stack files manually rather than the stack files on the repo, and update the documents to explain it. However, users who are new to sandstorm should be able to get the tutorial running straight away, instead of spending time customizing the php and the os versions IMHO. I’ll have a look at the documents and try to find a tradeoff before merging anything.

xet7:
Some versions of some web frameworks don't support PHP 7.x yet, PHP version should be selectable.

But in general latest versions of web frameworks support PHP 7.x, and for example latest MSSQL drivers are only available for PHP 7.x, so I don't know yet how many older apps require older PHP version than 7.x.

ocdtrekkie:
There are only a handful of Sandstorm apps using PHP, we probably could check whether or not the current versions of those apps support PHP 7. I'd definitely like to have a php5 stack and a php7 stack, but I suppose it's on @kentonv whether or not we should keep the old jessie64 version, just to support php5.

kentonv:
@Rivierg Thanks for helping us stay on top of things. I personally don't know much about vagrant-spk internals so having contributors update things is really helpful. Please let me know if/when you think this should be merged.

ocdtrekkie:
@kentonv My understanding is that #211 should be merged, because being on Stretch requires we switch to PHP 7 for the git master to be "basically functional" for the PHP stack.

Note that WordPress, MediaWiki, Firefly III, Matomo (nee Piwik), TinyTinyRSS, Lychee, and Roundcube all appear to either support or _explicitly recommend_ PHP 7.x. Which is all of the PHP apps I could come up with on Sandstorm. We can help someone customize vagrant-spk back to jessie and PHP 5 if they really need it.

PR #214, opened 2018-03-15
Update lesp stack for PHP 7

ocdtrekkie:
This update is **completely untested**. However, I feel it should be merged. Essentially, I mirrored @Rivierg's modifications to the lemp stack, which is probably 9/10ths of what is needed to ensure lesp also works on Debian Stretch. All package renames and location changes are identical to the changes in #211, except for php7.0-sqlite3, which I had to look up the name of, since they added the SQLite version to the package name according to https://packages.debian.org/stretch/php7.0-sqlite3

### DIFF
--- a/stacks/lesp/launcher.sh
+++ b/stacks/lesp/launcher.sh
@@ -2,7 +2,7 @@
 
 # Create a bunch of folders under the clean /var that php and nginx expect to exist
 mkdir -p /var/lib/nginx
-mkdir -p /var/lib/php5/sessions
+mkdir -p /var/lib/php/sessions
 mkdir -p /var/log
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
I tested and this has some of the same issues as #220. The existing lesp stack is also broken because Stretch doesn't offer up PHP 5.

ocdtrekkie:
Okay, I believe this is now fixed. @JC5 had an extra sed statement that fit the bill nicely for what was apparently missing. I was able to run a PHP script and connect to an empty SQLite with it, and display PHPInfo.

JC5:
:+1:

PR #216, opened 2018-04-17
Update sources before installing curl

### LINKED ISSUE #215: vagrant-spk fails to install curl because of a stale package index
Hi

https://github.com/sandstorm-io/vagrant-spk/blob/08fc4dbdd2847ad59e3fea25fa480247ff6979b0/vagrant-spk#L185

needs to be

`apt-get update`
`apt-get install -y curl`

(I'm following the tutorial)

ocdtrekkie:
Fixes #215.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -182,6 +182,7 @@ echo localhost > /etc/hostname
 hostname localhost
 
 # Install curl that is needed below.
+apt-get update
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I've confirmed @QJKX's report, and that his fix works.

PR #222, opened 2018-04-25
Fix LEMP stack, install MySQL

### LINKED ISSUE #220: PHP sample app/tutorial does not work
Everything with the tutorial and PHP sample app works out great... except the app doesn't start. I'm guessing between moving to Stretch and PHP 7, things broke. Below is the debug log I'm working from.

https://docs.sandstorm.io/en/latest/vagrant-spk/packaging-tutorial/
https://github.com/sandstorm-io/php-app-to-package-for-sandstorm

Perhaps @Rivierg might have some input here?

> Debug log: Untitled Example App instance
...** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns
chown: invalid user: 'mysql'
Cannot change ownership of the database directories to the 'mysql'
user.  Check that you have the necessary permissions and try again.
waiting for mysql to be available at /var/run/mysqld/mysqld.sock
2018-04-23  0:45:41 140581096079936 [Note] /usr/sbin/mysqld (mysqld 10.1.26-MariaDB-0+deb9u1) starting as process 30 ...
/usr/sbin/mysqld: One can only use the --user switch if running as root
[23-Apr-2018 00:45:41] ERROR: unable to bind listening socket for address '/run/php/php7.0-fpm.sock': Read-only file system (30)
[23-Apr-2018 00:45:41] ERROR: FPM initialization failed
waiting for mysql to be available at /var/run/mysqld/mysqld.sock
waiting for mysql to be available at /var/run/mysqld/mysqld.sock
waiting for mysql to be available at /var/run/mysqld/mysqld.sock
waiting for mysql to be available at /var/run/mysqld/mysqld.sock
...

ocdtrekkie:
So Debian Stretch installs MariaDB instead of MySQL. That breaks vagrant-spk, so for now, this adds repo.mysql.com as an apt source and gets MySQL proper. MySQL expected an additional directory in var (seems to to `secure_file_priv`), and indicated that the old way of initalizing the database was deprecated, so we're going to try changing that too.

Additionally, this fixes PHP 7's sock file location as with the update-lesp patch.

Fixes #220.

### DIFF
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -21,7 +21,7 @@ mkdir -p /var/run/mysqld
 HOME=/etc/mysql /usr/sbin/mysqld --initialize
 
 # Spawn mysqld, php
-HOME=/etc/mysql /usr/sbin/mysqld &
+HOME=/etc/mysql /usr/sbin/mysqld --skip-grant-tables &
 /usr/sbin/php-fpm7.0 --nodaemonize --fpm-config /etc/php/7.0/fpm/php-fpm.conf &
 # Wait until mysql and php have bound their sockets, indicating readiness
 while [ ! -e /var/run/mysqld/mysqld.sock ] ; do
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -2,6 +2,7 @@
 
 # Create a bunch of folders under the clean /var that php, nginx, and mysql expect to exist
 mkdir -p /var/lib/mysql
+mkdir -p /var/lib/mysql-files
 mkdir -p /var/lib/nginx
 mkdir -p /var/lib/php/sessions
 mkdir -p /var/log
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
Oh, shoot, I forgot I had more things to fix with MySQL. Anyways, this is a lot of progress and the explanation for it.

I figured out the secure_file_priv bit relating to the new folder here: https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_secure_file_priv

ocdtrekkie:
Okay, so... I found a fix, but I have a security question attached, @kentonv.

For some reason, root with no password (which is how the existing stack works) isn't being accepted by the updated stack. But if I launch `mysqld` with `--skip-grant-tables`, which is meant for resetting the password, and essentially disables all user security... it works.

In the context of "inside a Sandstorm grain", is "just shutting off authentication" a valid way to handle MySQL? Presumably if someone had a reason to do otherwise, they'd modify their stack anyways?

-----
Research:

Section B.5.3.2.2 here: https://dev.mysql.com/doc/refman/5.7/en/resetting-permissions.html

kentonv:
Yeah, MySQL authentication seems pointless inside a Sandstorm grain, since external clients cannot connect anyway and a grain is not multi-tenant.

PR #223, opened 2018-04-26
Update version number in Windows installer

### LINKED ISSUE #208: Cut a new release
I'd like to be able to use the new Go stack on a Windows host. :)

ocdtrekkie:
Set for next Sandstorm release version (somewhat optimistically). But either way it'll be more accurate than a date in 2015, and give some clear indication if a user has installed a modern vagrant-spk.

Fixes #208.

### DIFF
--- a/windows-support/windows-installer.iss
+++ b/windows-support/windows-installer.iss
@@ -2,7 +2,7 @@
 ; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
 
 #define MyAppName "vagrant-spk"
-#define MyAppVersion "2015-07-24"
+#define MyAppVersion "v0.233"
 #define MyAppPublisher "Sandstorm Development Group, Inc."
 #define MyAppURL "https://docs.sandstorm.io/"
 #define MyAppExeName "vagrant-spk.exe"


kentonv:
Maybe we should stick to using a date here, since it was a date in the past? I'm not sure it really makes sense to tie vagrant-spk release versions to Sandstorm release versions if we don't expect to release them together (which we don't, do we?).

ocdtrekkie:
vagrant-spk was never versioned by date, and this date was never correlated with any release identifiers for Linux/Mac users whatsoever, or even updated when the Windows release was updated.

vagrant-spk's always been versioned to whatever the current Sandstorm release is. I'm not super attached to that personally, mind you, I recall once before, Asheesh needing to wait for Sandstorm's version number to increment so something could be released, and that's arguably a potentially worse problem with the slower release cadence of Sandstorm today.

But on the other hand, vagrant-spk directly calls Sandstorm spk functions and installs Sandstorm in it's VM, and arguably, changes to Sandstorm may necessitate changes in vagrant-spk, so there's some sense in it being correlated somewhat.

I'd argue this change is correct for now. If changes are to be made to vagrant-spk's versioning structure, that should take the release.sh script into account. And as far as when to do that, I'd be inclined to get all of vagrant-spk up to a smooth running state and then call it 1.0.

PR #228, opened 2018-05-01
Update uwsgi stack for Stretch

ocdtrekkie:
This PR isn't complete yet, but is currently "better than what currently happens". It updates the MySQL install behavior from my PHP PR (#222), so that we don't install MariaDB.

Current issue using python test app:

> /opt/app/env exists, moving on
/opt/app/.sandstorm/build.sh: line 11: /opt/app/env/bin/pip: No such file or directory
Connection to 127.0.0.1 closed.
[31mCommand failed with a non-zero exit status (127).[0m

I suspect pip needs to be installed somehow, as it is indeed not in /opt/app/env/bin.

### DIFF
--- a/stacks/uwsgi/launcher.sh
+++ b/stacks/uwsgi/launcher.sh
@@ -17,10 +17,11 @@ mkdir -p /var/run/mysqld
 UWSGI_SOCKET_FILE=/var/run/uwsgi.sock
 
 # Ensure mysql tables created
-HOME=/etc/mysql /usr/bin/mysql_install_db --force
+# HOME=/etc/mysql /usr/bin/mysql_install_db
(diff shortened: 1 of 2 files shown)


kentonv:
To be clear, do you want me to merge this as-is or do you want me to wait until it is more complete?

ocdtrekkie:
At some point I intend to finish this, but it definitely does not work without this change as vagrant-spk's adjustments to MySQL do not work on MariaDB. This is arguably "less broken" than what is currently in master, and gets someone closer to a working build should they try to use it.

tl;dr: Master is broken and this is less broken, no reason not to merge it for now, saves me having to re-add it manually when testing.

PR #231, opened 2018-05-23
Update Meteor stack to 1.6.1.1

### LINKED ISSUE #152: vagrant-spk needs to point to newer meteor-spk bundle to support Meteor 1.3
Project is Meteor 1.3.1

After making with vagrant-spk dev:

When attempting to load a grain, nothing comes up but a loady spinny gif.

Going directly to the iframe URL results in a pending request to load it, which comes back as a failed after 2 minutes, then attempts to load again, then about after 5 minutes returns a 504.

The project was not building in vagrant-spk initially because a node package, fibers, said it needed to be built. I was able to run after installing g++ manually on box.

I still get these warnings, though:
npm WARN package.json meteor-dev-bundle@0.0.0 No description
npm WARN package.json meteor-dev-bundle@0.0.0 No repository field.
npm WARN package.json meteor-dev-bundle@0.0.0 No README data

After it builds and runs, I logged in with vagrant-spk ssh to test some things. Don't know if this will help.

vagrant@localhost:/var$ curl http://f98c7611a5f51619e252dd432a8a42d1.local.sandstorm.io:6080/_sandstorm-init?sessionid=07278101882155671c1e4f055235db4bac146cdb91d37a73a23a32158d1e62aa&path=/
[1] 24739

vagrant@localhost:/var$ curl http://f98c7611a5f51619e252dd432a8a42d1.local.sandstorm.io:6080/
Error: Unauthorized [403]
    at Proxy.makeContext (server/proxy.js:1512:15)
    at server/proxy.js:1711:20
    at C (/node_modules/es6-promise.js:12:57)
    at /node_modules/es6-promise.js:17:252
    at t (/node_modules/es6-promise.js:9:473)
    at process._tickCallback (node.js:448:13)[1]+  Done                    curl http://f98c7611a5f51619e252dd432a8a42d1.local.sandstorm.io:6080/_sandstorm-init?sessionid=07278101882155671c1e4f055235db4bac146cdb91d37a73a23a32158d1e62aa

### LINKED ISSUE #193: Offer a Meteor 1.4-compatible vagrant-spk platform stack
Context:
- If you try to package an app that uses Meteor 1.4.x with vagrant-spk's Meteor platform stack, you will run into mysterious build problems.
- This is due to version incompatibilities between the nodejs version bundled with Meteor pre-1.4 and the node version bundled with Meteor 1.4+.
- A while ago, @zarvox and I discussed this, and the best solution we could come up with was to have two different platform stacks -- one for Meteor 1.4+ and one for Meteor pre-1.4.

Sub-problems to consider:
- vagrant-spk should properly package a Meteor apps that has no `.sandstorm/`. This may involve looking at  `.meteor/release` and printing an error message, or auto-switching which `.sandstorm/` it generates.
- vagrant-spk should provide reasonable advice for apps that used to Meteor 1.3 and now are using Meteor 1.4. This may involve looking at  `.meteor/release` and printing an link to docs.

ocdtrekkie:
Heavily based on @aruntk's work (see https://github.com/CommonGarden/Grow-IoT/commit/3f10818a3fa861127f113711099914c7157c7e47), I am trying to mark him as a co-author on the commit.

Fixes #152 and fixes #193 (and several issues on other repositories).

### DIFF
--- a/stacks/meteor/build.sh
+++ b/stacks/meteor/build.sh
@@ -3,12 +3,19 @@ set -euo pipefail
 
 # Make meteor bundle
 
-METEOR_WAREHOUSE_DIR="${METEOR_WAREHOUSE_DIR:-/home/vagrant/.meteor}"
-METEOR_DEV_BUNDLE=$(dirname $(readlink -f "$METEOR_WAREHOUSE_DIR/meteor"))/dev_bundle
+export NODE_ENV=production
+sudo chown vagrant:vagrant /home/vagrant -R
 
 cd /opt/app
+meteor npm install --production
 meteor build --directory /home/vagrant/
-(cd /home/vagrant/bundle/programs/server && "$METEOR_DEV_BUNDLE/bin/npm" install)
+# Use npm and node from the Meteor dev bundle to install the bundle's dependencies.
+TOOL_VERSION=$(meteor show --ejson $(<.meteor/release) | grep '^ *"tool":' |
+    sed -re 's/^.*"(meteor-tool@[^"]*)".*$/\1/g')
+TOOLDIR=$(echo $TOOL_VERSION | tr @ /)
+PATH=$HOME/.meteor/packages/$TOOLDIR/mt-os.linux.x86_64/dev_bundle/bin:$PATH
+cd /home/vagrant/bundle/programs/server
+npm install --production
 
 # Copy our launcher script into the bundle so the grain can start up.
 mkdir -p /home/vagrant/bundle/opt/app/.sandstorm/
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
Note that this stack expresses the opinion that Meteor-based applications on Sandstorm tend to already support this version of Meteor, and that new packages should be assumed to run on it as well. We can always assist people if they need advice on packaging with old versions of platform stacks.

Stack scripts are only used when running a new `setupvm`, it will not break existing working configurations.

kentonv:
Thanks!

PR #240, opened 2018-08-06
Fixed mingw total ram calculation command

spollard:
Removed an extra "}" from the command


ocdtrekkie:
I can definitely see an error when I run the command as it was, and don't see the error when I run your corrected command, but I also, interestingly, don't actually get a response at all from PowerShell when I run that. I actually think you got it wrong on that bracket being "extra". The nice thing about PowerShell being PowerShell, is that a command like this can quickly be tested in the prompt. :)

The original command returns an error when you run it:
`Get-CimInstance -class cim_physicalmemory | % $_.Capacity}`

This (the command in your PR) returns nothing.
`Get-CimInstance -class cim_physicalmemory | % $_.Capacity`

This returns the amount of RAM on my PC.
`Get-CimInstance -class cim_physicalmemory | % {$_.Capacity}`

ocdtrekkie:
Actually, more interestingly, I think all of these are broken. The result I get is:

`8589934592
8589934592`

Which is weird, I have 16 GB of RAM, it is returning the RAM from each RAM chip separately, I presume. Since the program treats the output like an integer, this is never going to work right, I think.

Meanwhile `[math]::Round((Get-WmiObject -Class Win32_ComputerSystem).TotalPhysicalMemory)`
returns `17117069312`

PR #241, opened 2018-08-15
Fix lookup of total memory on Windows machines

ocdtrekkie:
Credit where credit's due, this problem was discovered by @spollard and the related analysis is in #240. However, none of the versions of the old command returned a single integer representing the total RAM. Either no result was returned or two results, each containing half the RAM, was returned as a table. This alternative returns a single integer value for the total system memory, and should actually work here.

When @zarvox wrote this, he wrapped it in a rescue block to handle computers without PowerShell, but my guess is that this conveniently also caused it not to fail when his PowerShell was not returning valid responses as well.

Closes #240.


ocdtrekkie:
It should be noted I also have no real way to test this PR in execution since vagrant-spk on Windows needs to be compiled, but the old PowerShell did not return a valid number, and this one does.

Issue #233, opened 2018-05-28
To Do: Update Node stack(s)


ocdtrekkie:
It looks like currently we have `node` which uses Node 6.x, and `node7` which uses Node 7.x.

What needs to be realistically determined is what the current state is of tendencies of apps to run on versions of Node, because I'd rather have just one stack for it, if we aren't living in a Python 2 vs. Python 3 sort of world here, where there's long term need for the legacy version to be supported.

According to https://github.com/nodejs/Release#release-schedule Node 6 is in maintenance until April 2019, Node 8 is being actively developed until April 2019 and in maintenance well past that, and Node 10 is now the current version.

PR #242, opened 2019-09-11
fix: update to node 10 LTS and drop Node 6 & 7

curbengh:
Closes https://github.com/sandstorm-io/vagrant-spk/issues/233

also remove apt-key import as the NodeSource installation script already [includes](https://github.com/nodesource/distributions/blob/0114ee84afe76633c9cd28958c2d09cceee84f49/deb/setup_10.x#L284) the step.

https://github.com/nodejs/Release#end-of-life-releases
Node 6 EOL 2019-04-30
Node 7 EOL 2017-06-30 (!)

Node 8 will EOL 2019-12-31
Node 10 is supported until April 2021.

`node` folder could be renamed to `node6` to support legacy apps (Edit: renamed).

---

Alternatively, Sandstorm can opt for the versioning used by [Docker](https://hub.docker.com/_/node) (as commonly found in most CI). Meaning `node` refers to latest stable (currently v12, soon v13), whereas `nodelts` refers to v10 lts.

But that means the stack needs to be updated every 6 months. Using the latest node (in Sandstorm) has no significant benefit since Node apps usually (and should) support lts.


curbengh:
I raise this PR because recently I'm trying to package [Hexo](https://github.com/hexojs/hexo) as a Sandstorm app. Hexo is going to drop Node 6 soon, so it wouldn't run on the `node` stack. While I'm aware that I can always go for the `diy` route, but why `diy` when `node` stack is available.

---

It's not directly related to the PR, I want to inquire about the OS used in Sandstorm. The current node stack shows Debian Jessie. Then I saw
https://github.com/sandstorm-io/vagrant-spk/blob/faa13cc2aef167d33db89cff14c0c0d2acc936fb/stacks/node/setup.sh#L13
so how do I choose Ubuntu over Debian? But not including [`export DEBIAN_FRONTEND=noninteractive`](https://github.com/sandstorm-io/vagrant-spk/blob/faa13cc2aef167d33db89cff14c0c0d2acc936fb/stacks/node/setup.sh#L15) line?

ocdtrekkie:
vagrant-spk uses Debian Stretch, not Jessie. I think the Node stack was just never updated. You'll see here in vagrant-spk that the Debian Stretch image is selected during a vagrant-spk init: https://github.com/sandstorm-io/vagrant-spk/blob/master/vagrant-spk#L73

When we switched from Jessie to Stretch, we had to update each and every platform stack to properly work with the newer packages and the like, so changing what flavor of Linux vagrant-spk uses is very non-trivial, there is no configurable option in vagrant-spk to use Ubuntu instead, although it's not inherently impossible to edit vagrant-spk to do so. Note that Sandstorm does not care what flavor of Linux you use, it's just that vagrant-spk is using Debian.

When you package an app with vagrant-spk, it creates a virtual machine with Debian, installs Sandstorm in it, and lets you work on developing and packaging your app. The resulting app may include some dependencies as shipped by the flavor of Debian you used, but will run on whatever Linux kernel your Sandstorm is installed on. Note that if you use the spk tool, instead of vagrant-spk, you can happily build a Sandstorm app using whatever flavor of Linux you installed Sandstorm on as well. There's also a newer docker-spk tool, and the example app for it actually builds off of Alpine Linux.

ocdtrekkie:
Renaming the existing node stack to node6 would be a good choice for now, though already-packaged apps should effectively have all of those files already, so there shouldn't be a whole lot of harm in purging the old node stacks at some point here.

I think node and nodelts seems like a good strategy for the future. If you don't need node12 at this point, and think most packages will work with nodelts, perhaps we should rename the existing node stack, so people don't use it accidentally, create the nodelts stack for v10, and then not worry about making a new plain node stack unless someone needs it? Right now vagrant-spk doesn't get a lot of development love, so it's not going to be updated on a six month cycle, it's going to be updated on a "someone submitted a PR and bugged Kenton to push the updates out" cycle.

curbengh:
> vagrant-spk uses Debian Stretch, not Jessie.

Seems like the docs doesn't mention it, I expected "Build Environment" section like how [Travis](https://docs.travis-ci.com/user/reference/overview/) document it. I'll submit a PR for this.

> Right now vagrant-spk doesn't get a lot of development love, so it's not going to be updated on a six month cycle

As long I'm still active on Hexo, I don't mind PR regularly, but yeah, someone needs to merge it.

In that case, I suggest `node` to use v10 lts and rename current `node` to `node6`. The stacks are used for production deployment, so using lts is more appropriate. Since latest node is usually only used for testing, user can opt to `diy` instead.

ocdtrekkie:
Assuming you've tested that this works (I don't have time to right now), I'm pretty happy with this PR. I like how the new version of `node6` uses Node's own install script which is release-independent, rather than hardcoding our build (and in this case, it was using jessie because the stack hadn't been updated).

Furthermore, this clearly defines and decides what version the `node` stack should be running going forwards (the current LTS), while leaving the legacy stack intact for people who may wish to use or refer back to it.

ocdtrekkie:
I am looking to merge this and build a vagrant-spk 1.0 release in the very immediate future. Just an FYI.

ocdtrekkie:
Did some rudimentary testing, I'm happy with it. The old ones are 100% broken anyways. :)

PR #252, opened 2020-01-31
Fix some bitrot in enter_grain

zenhack:
with #![feature(alloc_system)] I get an error from the rust toolchain
that the feature has been removed. We don't appear to actually be using
this anyway, so let's just remove it.


zenhack:
Maybe hold off on this one until we decide whether to just switch over to using nsenter.

ocdtrekkie:
@zenhack Agreed on that.

paulproteus:
I'm excited to see the bitrot being fixed, and also excited to see the
possibility that we'll delete this code. :)

For what it's worth, some of the flags I was using were due to my desire to
keep the output binary file as small as possible. Maybe this was one of
them.

zenhack:
Hm, the binary on master is >600K vs this one at 200K, so if that's the case, the compiler has gotten better by a margin that swallows those savings pretty thoroughly. We still shouldn't commit it to the repo though...

ocdtrekkie:
Since #253 is concerned about excess commits of binary files, if you fix the enter-grain binary, can we get both this fix and that fix in a single commit, so we are only adding one copy of the binary?

zenhack:
Sure, I'll --amend it onto this one. We can maybe do the PyPI packaging soonish so we don't have to worry about this again.

zenhack:
@ocdtrekkie, updated. should work now.

ocdtrekkie:
I tested this locally and I'm happy with it.

Issue #244, opened 2020-01-21
Cleanup: Function that recommends sandstorm/jessie64


ocdtrekkie:
So this function mostly just checks if a specific version of the jessie box is in use and complains about it. vagrant-spk now defaults to stretch, it's actually the box this function that recommends which no longer exists at all and doesn't work, and we've settled on official contrib boxes working fine.

https://github.com/sandstorm-io/vagrant-spk/blob/master/vagrant-spk#L320

Honestly we can probably mostly get rid of this? If anything maybe we should check for uses of sandstorm/jessie64 and recommend they change it to debian/contrib-jessie64... or stretch64.

zenhack:
Yeah I'd be in favor of just deleting that whole swath of code.

ocdtrekkie:
I might look at doing the thing where we detect if they were trying to use the sandstorm/ box, just because it's been such a prevalent complaint when people try to rebuild apps, and I don't see it not being a question soon. But even that would be shorter/cleaner than this.

PR #257, opened 2020-02-04
Add upgradevm command

### LINKED ISSUE #245: Need a way for vagrant-spk to update scripts
Regularly we have people trying to update old apps hitting the 404 for the old sandstorm/jessie64 box. vagrant-spk doesn't use that box anymore, but the scripts in the .sandstorm directory for random old apps don't know that.

My advice is generally to do a new setupvm in another folder, and grab the Vagrantfile and global-setup.sh from it and overwrite what's in the app directory. I think we should have a command like maybe `vagrant-spk updatevm` which overwrites those two files in your existing app's .sandstorm directory with whatever vagrant-spk generates for new apps.

I do not think we can do this for the stack files because apps customize them too much. But some will not break at all (especially if they used the diy stack). Apps shouldn't ever be tampering with the Vagrantfile or global-setup.sh, so I think it would be relatively safe to do this, but if we create this command we should also add comments to these two files to specify explicitly that people should not tamper with them, and maybe have a warning step when executing it that it will overwrite those files.

ocdtrekkie:
Fixes #245 

Super simple for now. It grabs the same code as setupvm but just for Vagrantfile and global-setup.sh, which shouldn't generally be modified by app packagers.

As a test, I grabbed the GitWeb source prior to @abliss updates, because I knew he hit the 404 box issue. Went through the flow and failed to start the VM. I ran `vagrant-spk upgradevm` and then I was able to start up the VM. (Although I also got a 502 Bad Gateway because there was other related bitrot for GitWeb.)

Note that currently neither `setupvm` nor `upgradevm` has any sort of protection to prevent you from clobbering what you have. (This is worse in setupvm's case, because it can mess up your app build.) Though hopefully people are committing to GitHub or have original sources to fall back on. But we can do better here. We should also put comments at the top of both files we upgrade this way to further discourage user customization to make this command 'safer'.

Also, I'm not yet detecting the box that caused GitWeb to 404, and we're going to amend the detection for that to do so.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -53,6 +53,8 @@ CODE_DIR = os.path.dirname(EXECUTABLE_PATH)
 VAGRANTFILE_CONTENTS = r"""# -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+# CAUTION: DO NOT MAKE CHANGES TO THIS FILE. The vagrant-spk upgradevm process will overwrite it.
+
 # Guess at a reasonable name for the VM based on the folder vagrant-spk is
 # run from.  The timestamp is there to avoid conflicts if you have multiple
 # folders with the same name.
@@ -179,6 +181,9 @@ GITIGNORE_CONTENTS = r"""
 
 GLOBAL_SETUP_SCRIPT = r"""#!/bin/bash
 set -euo pipefail
+
+# CAUTION: DO NOT MAKE CHANGES TO THIS FILE. The vagrant-spk upgradevm process will overwrite it.
+# App-specific setup should be done in the setup.sh file.
 
 # Set options for curl. Since we only want to show errors from these curl commands, we also use
 # 'cat' to buffer the output; for more information:
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -323,51 +323,29 @@ def call_vagrant_command(sandstorm_dir, *command_args):
         sys.exit(ANSI_RED + msg + ANSI_RESET)
 
 def ensure_working_vboxsf_in_base_box(sandstorm_dir):
-    # If the .sandstorm/Vagrantfile refers to debian/jessie64, and does not pin it to a particular
-    # version, and the user has version 8.2.2 available on their system, then we abort the
-    # "vagrant-spk up" process because it will not work.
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
This now also fixes #244 

There was a lot of complicated logic in deciding not to flip out if someone pinned an old enough jessie64 box. I decided we don't want to continue to support super old pinned versions of jessie64 boxes anyways, and scrapped all of it. Though I left instructions to update to a supportable contrib-jessie64 box if someone really wants to get a package working on Jessie, we'd prefer people upgradevm to clear out any and all Vagrantfile-related bitrot, of which there is plenty since that era of vagrant-spk.

ocdtrekkie:
Now when downloading the pre-Adam-updated GitWeb, instead of getting a 404 for trying to download the box that doesn't exist, we tell people they should upgradevm or switch boxes to contrib-jessie64. After running upgradevm, the correct versions are placed in the directory and the VM can correctly boot up.

abliss:
Cool!

vagrant-spk
```diff
@@ -318,51 +323,29 @@ def call_vagrant_command(sandstorm_dir, *command_args):
...
-
-Open .sandstorm/Vagrantfile and find this line:
-
-  config.vm.box = "debian/jessie64"
-
-Change it to:
+This package was set to use a VM no longer compatible with vagrant-spk. You
+can either upgrade to the latest VM supported by vagrant-spk by running
+"vagrant-spk upgradevm" or manually update the package to use a compatible
+release of Debian Jessie.
```
zenhack: Vagrantfile says it uses stretch now, not jessie? Probably needs updating elsewhere too.

ocdtrekkie: This is intentional. You can _either_ use `upgradevm` to use the latest (currently: stretch) or we explain how to use the "correct" version of Jessie, if your app is really dependent on it being that and you want to make it work with Jessie.

ocdtrekkie: Definitely intending to add confirmation: https://github.com/sandstorm-io/vagrant-spk/issues/258

setupvm will already happily burn everything to the ground if you rerun it on an existing app.

zenhack: Aha, makes sense.

PR #261, opened 2020-02-13
Set release.sh to version 1.0 manually

ocdtrekkie:
I mostly stripped out the logic for determining the version, and added the variables for SemVer. For most intents and purposes, I am looking at major.minor, but I have added the ability to set a patch as necessary, upon which it will show the patch number.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -34,34 +34,17 @@ function assert_git_state_is_clean() {
 }
 
 function get_release_name() {
-  # We use the current Sandstorm release number as vagrant-spk release number for simplicity. This
-  # does mean it might be hard to hae two vagrant-spk releases in a short time period.
-  #
-  # As for what these version numbers *mean* and how this gets calculated, see the sandstorm
-  # repository's release.sh.
-  local CHANNEL=dev
+  BUILD_MAJOR = 1
+  BUILD_MINOR = 0
+  BUILD_PATCH = 0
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Note this is completely untested at the moment, and just "seems right".

release.sh
```diff
@@ -34,34 +34,17 @@ function assert_git_state_is_clean() {
...
-    echo "ERROR: $CHANNEL has already moved past this branch!" >&2
-    echo "  I refuse to replace it with an older branch." >&2
-    exit 1
+  BUILD_MAJOR = 1
+  BUILD_MINOR = 0
+  BUILD_PATCH = 0
+
+  # DISPLAY_VERSION gets used in the git tag description
+  DISPLAY_VERSION="${BUILD_MAJOR}.${BUILD_MINOR}"
+  if (( BUILD_PATCH > 0 )); then
```
zenhack: You'll want to actually expand the `$BUILD_PATCH` variable, otherwise it's just comparing against the *string* `BUILD_PATCH`.

ocdtrekkie:
I think I came up with where I was from "cut out the things that explicitly do not build the version I want, and insert my own". Entirely correct that all of that is unnecessary though.

Dumb bash scripting comes from figuring out what I want to do when on a machine that doesn't have bash.

PR #262, opened 2020-02-14
WIP: Verify all version numbers (DO NOT MERGE)

ocdtrekkie:
Checks the version number in the script itself and the Windows install configuration.

I've tested each one prior to updating the version number at the location in question and after to verify that it fails/passes each check correctly respectively. It is very sad if we push a release which reports the wrong version either when checked by command (vagrant-spk --version) or in the Windows installer, and it has happened before.

This method is a little bit fragile in that it picks a precise line number that it should be on, and if we amend the top of either file we may need to fix this, but there is not much functional code above either line, so I suspect this will not be a frequent problem. It should be good enough for now.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -51,6 +51,26 @@ function assert_changelog_present() {
   fi
 }
 
+function assert_version_updated() {
+  # Verify that the vagrant-spk version has been updated.
+  EXPECTED_SCRIPT_VERSION="__version__ = \"$TAG_NAME\""
+  if [ "$(head -n 26 vagrant-spk | tail -n 1)" != "$EXPECTED_SCRIPT_VERSION" ]; then
+    echo "vagrant-spk version not updated. 26th line should be:" >&2
+    echo "$EXPECTED_SCRIPT_VERSION" >&2
+    exit 1
+  fi
+}
+
+function assert_winstall_updated() {
+  # Verify that the Windows installer version has been updated.
+  EXPECTED_WINSTALL_VERSION="#define MyAppVersion \"$TAG_NAME\""
+  if [ "$(head -n 5 windows-support/windows-installer.iss | tail -n 1)" != "$EXPECTED_WINSTALL_VERSION" ]; then
+    echo "Windows installer version not updated. 5th line should be:" >&2
+    echo "$EXPECTED_WINSTALL_VERSION" >&2
+    exit 1
+  fi
+}
+
 function build_windows_exe() {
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -51,6 +51,26 @@ function assert_changelog_present() {
   fi
 }
 
+function assert_version_updated() {
+  # Verify that the vagrant-spk version has been updated.
+  EXPECTED_SCRIPT_VERSION="__version__ = \"$TAG_NAME\""
+  if [ "$(head -n 26 vagrant-spk | tail -n 1)" != "$EXPECTED_SCRIPT_VERSION" ]; then
```
zenhack: Could we not just run `vagrant-spk --version` here?

ocdtrekkie: Yeah, we can.

ocdtrekkie:
I think it's the windows-installer.iss file I'm most concerned about. I'm not sure if there's either a way to cause it to execute a vagrant-spk --version query to set it. Or a good way to query to check it other than the method deemed above. I'd have to spend some time investigating Inno Setup, and I don't know how much time I want to invest in that: Bumping the version is cheap and hard to break.

zenhack:
It looks like it has an #include directive (basically c-preprocessor it
looks like...). Maybe just generate a file with the one #define via the
makefile, something like:

vagrant-spk --version | awk '{print "#define MyAppVersion \"" $2 "\""}' > version.iss


Quoting Jacob Weisz (2020-02-14 12:06:12)

ocdtrekkie:
That sounds sensible and I agree on all points. Though I would want to play with testing that a bit. (I actually haven't even gotten to the "run Inno Setup" step of prepping this release, so I'm expecting the fun has not yet ended.)

I think as an improvement that is important prior to GitHub Actions doing releases, but one that doesn't functionally impact the release, I am going to work on this one after the current release.

zenhack:
Sounds good to me.

ocdtrekkie:
Referencing this thought for later: Even with your change to remove manually updating the Windows installer config, we're still updating: release.sh, vagrant-spk --version, and the changelog.

So the correct improvement is probably also to change get_release_name in release.sh again to use vagrant-spk --version for TAG_NAME, create DISPLAY_NAME by dropping the first character off. Which drops us to two places to update the version: The vagrant-spk script itself and the changelog.

Then instead of changes to release.sh firing the Action, we have it fire on changes to the changelog (which seems pleasant), where it still verifies the most recent changelog entry carries today's date, and we'll still check against vagrant-spk --version to make sure it matches.

PR #263, opened 2020-02-17
Add server link

zdb999:
Add a link to the server launched on when using the command "vagrant-spk vm up." This makes it easier to launch Sandstorm in the browser without rechecking the documentation.


ocdtrekkie:
I like this change a lot, and have been trying to get `spk` in Sandstorm updated to do this dynamically (https://github.com/sandstorm-io/sandstorm/issues/3215), particularly when you type `vagrant-spk dev` or `spk dev`. But this, at least, a useful option for now.

However, as of #260, this is actually wrong, because new vagrant-spk VMs use version 6090, not 6080. (This change is part of why I agree it's important for it to tell you where your server is running.) I'm happy to merge a PR adding this as long as it specifies the correct (new) port.

zdb999:
I just changed the port to 6090 to reflect the changes in #260.

ocdtrekkie:
Thanks!

PR #266, opened 2020-03-16
Unpin box and apt upgrade

### LINKED ISSUE #248: Unpin box and include apt-get upgrade
vagrant-spk should not strongly push using a static box version and also fail to run apt-get, meaning old, potentially buggy or insecure binaries are packaged with apps by default. With the contrib- boxes maintaining vboxsf support, there's no good reason to pin the box to an old version.

ocdtrekkie:
Fixes #248 

I tested this with diy stack and the VM came up alright. I'll probably do a wholesome test with the Python test app before I merge it. I found grub-pc held up apt-get upgrade, and from a Sandstorm package building standpoint, we hardly care if grub is up to date on the VM, so I held it.

Note that we should document `vagrant-spk vm box update` somewhere for people who currently use vagrant-spk and have the old box downloaded already. But realistically this change should mostly update your box up to the same level anyways.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -73,7 +73,6 @@ end
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Base on the Sandstorm snapshots of the official Debian 9 (stretch) box with vboxsf support.
   config.vm.box = "debian/contrib-stretch64"
-  config.vm.box_version = "9.3.0"
   config.vm.post_up_message = "Your virtual server is running at http://local.sandstorm.io:6090."
 
 
@@ -191,8 +190,12 @@ CURL_OPTS="--silent --show-error"
 echo localhost > /etc/hostname
 hostname localhost
 
-# Install curl that is needed below.
+# Grub updates don't silent install well
+apt-mark hold grub-pc
 apt-get update
+apt-get upgrade -y
+
+# Install curl that is needed below.
 apt-get install -y curl
 
 # The following line copies stderr through stderr to cat without accidentally leaving it in the


zenhack:
It would be good to try to detect if the update changes the kernel version, and if so prompt the user to reboot the VM; IIRC I had some issues with fuse after doing this manually way back, and it seems like a likely common stumbling block if we do upgrades automatically.

ocdtrekkie:
@zenhack Wouldn't I have to do `apt-get dist-upgrade` for it to update the kernel? As we don't really care about the kernel from a Sandstorm packaging standpoint, I didn't use that command.

zenhack:
No, you can still get bug/security fixes and such for the kernel via regular `apt-get upgrade`.

ocdtrekkie:
My understanding was the kernel couldn't update without `dist-upgrade`. In that case, should we be using `dist-upgrade` here?

Next question: With regards to your fuse question, would this still be true if it was prior to installing Sandstorm? Because that's the only time this runs.

zenhack:
We should not use dist-upgrade; that's designed for updating to a new release of the *distro*.

Re: fuse: my guess is we could dodge the particular issue by doing `modprobe fuse` before upgrading; iirc the problem was that the after upgrading the on-disk kernel, the compatible version of the module was gone.

...but more generally it's a fragile state to be in, and I think at a minimum we should detect this and print a warning with "if you run into problems, try rebooting the VM"

ocdtrekkie:
That's likely outside my skillset/knowledge area. The other issue of printing a warning here, is that this is near the top of a very long automated process of setting up the whole VM with Sandstorm and all, I am not sure anyone would ever see/read such a warning.

Documentation I've found says `do-release-upgrade` is how you upgrade to the new release of a distro, and then `dist-upgrade` is just willing to "intelligently install or remove packages to resolve conflicts, and that the reason `dist-upgrade` upgrades the kernel and `upgrade` does not, is because upgrading the kernel requires adding a package.

If you are sure this might inadvertently update the kernel, would it be worthwhile to consider using apt-mark to hold the kernel as well?

zenhack:
> If you are sure this might inadvertently update the kernel, would it be worthwhile to consider using apt-mark to hold the kernel as well?

Let's do that in the short term. The holding back the kernel is unlikely to break anything for app developers, and we don't really need security fixes for this anyway. Maybe at some point I'll submit a patch to detect & warn.

ocdtrekkie:
As a `upgrade` vs. `dist-upgrade` test, I re-pinned the 9.3.0 box on my test VM, to ensure I had an "old" Stretch to start with. And let the changes to global-setup from this PR do their thing: It did not update the kernel, despite the kernel being able to be updated to the version used by the 9.12.0 box:

image

`apt-mark hold linux-image-amd64` should be harmless to add to this PR, but I don't think it'll actually do anything different. Not using dist-upgrade does seem to hold the kernel updates, because to update linux-image-amd64, it has to add a "new" package, linux-image-4.9.0-12-amd64, which `apt-get upgrade` cannot do.

zenhack:
I don't feel super strongly then; feel free to merge as is.

(also: can we make a point of not posting screenshots that are just text, using copy and paste instead? It's an a11y issue).

ocdtrekkie:
Noted regarding screenshots!

PR #275, opened 2020-08-12
Update the 'golang' stack.

zenhack:
Uses the latest version of the toolchain, and supports (only) Go
modules.

### DIFF
--- a/stacks/golang/build.sh
+++ b/stacks/golang/build.sh
@@ -7,6 +7,8 @@ if [ ! -e go.mod ]; then
 	printf "vagrant-spk's golang stack does not support older GOPATH\n"
 	printf "based projects. Try running:\n" >&2
 	printf "\n" >&2
+	printf "    vagrant-spk vm ssh\n" >&2
+	printf "    cd /opt/app\n" >&2
 	printf "    go mod init example.com/mypkg\n" >&2
 	exit 1
 fi
--- a/stacks/golang/build.sh
+++ b/stacks/golang/build.sh
@@ -1,19 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-# In order to import other packages in your own source tree, you'll have to
-# set this to the import url for the root of your repository. for example:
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
So, if I have a older project to package, say github.com/prologic/wiki, `go mod init github.com/prologic/wiki` should make it compatible with the upgraded Go stack?

zenhack:
Quoting Jacob Weisz (2020-08-12 09:45:36)

>    So, if I have a older project to package, say github.com/prologic/wiki,
>    go mod init github.com/prologic/wiki should make it compatible with the
>    upgraded Go stack?

Correct.

ocdtrekkie:
So, little bit of experimental testing here...

Outside of the VM, I have an old version of Go which doesn't know what `go mod` is. Some packagers (especially if just patching) might not have Go installed outside the VM at all. We might want to tell people to `vagrant-spk vm ssh` and then `cd /opt/app` before running mod init.

Then I got a blank go.mod file, which did nothing for me. A little searching found me `go mod tidy`, which populated the thing with more or less what I expected this Go project to have. Should we also recommend that there as well?

I kinda expected from the instructions that "go mod init" would upgrade a repo suitably for packaging, but we need to actually tell people to populate the go.mod file.

ocdtrekkie:
(I didn't try running vagrant-spk dev before running go mod tidy, so I guess that part of my comment is probably moot... Running it with a blank go.mod just seemed wrong? Anyways, I still think we may wish to tell people to run it in the VM, where the stack is installed.)

zenhack:
I updated the error message to suggest sshing into the VM first.

Per discussion on IRC, nothing extra is necessary. Go source files already import by URL, and you're supposed to change the import path if you break compatibility, so the extra information doesn't need to be specified and can be derived from the source files. The tools will update go.mod with the libraries they find references to, but in that case it's acting more like the auto-generated lock files you'll see for other package managers.

PR #277, opened 2020-08-31
Make vagrant-spk use new CSP when creating VMs

ocdtrekkie:
I've tested this change on ShareLatex on my Windows machine: It does indeed cause the vagrant-spk VM to use the correct config option. I placed it by another sandstorm.conf modification just prior to restarting the Sandstorm service, which seemed like a good spot.

This is to ensure new apps and app updates can easily be tested against sandstorm-io/sandstorm#3409. Track our progress on concluding this project at sandstorm-io/sandstorm#3424

vagrant-spk users should be using the latest commit of vagrant-spk, and may need to `vagrant-spk vm destroy`, `vagrant-spk upgradevm`, and then `vagrant-spk vm up` to take advantage of this for existing dev environments.


ocdtrekkie:
cc: @zenhack for review

PR #278, opened 2020-09-01
Bump the golang stack's version number.

zenhack:
Upstream just pushed out a patch release today, which fixes a security
vulnerability (I do not believe it affects anything that currently runs
in sandstorm).


ocdtrekkie:
Thanks! Perk of not using the distro version is we can be on the bleeding edge of these, I suppose!

Issue #247, opened 2020-01-28
vagrant-spk should be upgraded to Python 3


ocdtrekkie:
I don't think there's likely to be any blockers for doing this, vagrant-spk is fairly straightforward. And while this is a long/painful file to go through, I don't think it is exceptionally complicated, and I can probably tackle it.

ocdtrekkie:
I found this comment while exploring, and want to note it here since it's something we should account for/test when doing this: https://github.com/sandstorm-io/vagrant-spk/pull/145#discussion_r54626289

ocdtrekkie:
I ran a `pylint --py3k` on this, and got the following warnings:

no-absolute-import (W1618):
--
  | import missing `from __future__ import absolute_import` Used when an import is not accompanied by from __future__ import absolute_import (default behaviour in Python 3) This message can’t be emitted when using Python >= 3.0.

raw_input-builtin (W1609):
--
  | raw_input built-in referenced Used when the raw_input built-in function is referenced (missing from Python 3) This message can’t be emitted when using Python >= 3.0.
which is referring to https://github.com/sandstorm-io/vagrant-spk/blob/master/vagrant-spk#L798

ocdtrekkie:
https://stackoverflow.com/a/54712937/369990 seems like a way of fixing the raw_input issue without breaking Python 2.7 compatibility, which I think is probably preferable for now.

troyjfarrell:
I haven't read the whole script, but I'll make a point of using Python 3 exclusively with vagrant-spk for a while and file issues for any problems I encounter.

ocdtrekkie:
I am doing some testing with Python 3.8.2 on my Windows machine (using `python C:\...\vagrant-spk setupvm lemp` etc., which is wordy and tiring). It seems like it basically works.

ocdtrekkie:
So I no longer care about breaking python2 support, so, we should be able to fix raw_input more simply. As the case is, the only use of that function is part of the enter-grain flow. So I should fix that before merging #292 and make sure to test `vagrant-spk enter-grain`

ocdtrekkie:
It is possible there are still some hidden issues in vagrant-spk somewhere, but core functionality is tested, enter-grain is tested, which is usually the most unruly arm of it, and the best way to find out if we have more problems is probably to have people use it.

PR #285, opened 2021-02-03
Fix the setupvm command on Python 3.8

troyjfarrell:
This converts the str into bytes, which makes Python 3.8 happy.  Python 2.7 works before and after the change.


ocdtrekkie:
Do you know if there is anything else remaining to complete #247? Obviously we shouldn't tell people to use Python 2 anymore.

ocdtrekkie (approved):

PR #290, opened 2021-03-16
Use latest meteor-spk in Meteor stack

ocdtrekkie:
I haven't tested this, but on manual review, this looks like the only places this would need to be updated? I did check that the Cloudfront URL was still accurate for the Meteor download and such as well. Presumably we should have meteor-spk updated to 2.x at some point and then update this again.

### DIFF
--- a/stacks/meteor/setup.sh
+++ b/stacks/meteor/setup.sh
@@ -12,7 +12,7 @@ apt-get install -y build-essential git
 cd /opt/
 
 NODE_ENV=production
-PACKAGE=meteor-spk-0.4.1
+PACKAGE=meteor-spk-0.5.1
 PACKAGE_FILENAME="$PACKAGE.tar.xz"
 CACHE_TARGET="/host-dot-sandstorm/caches/${PACKAGE_FILENAME}"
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
cc: @gischer

zenhack:
This looks reasonable, though obviously someone should actually test it.

ocdtrekkie:
Agreed. I probably need to find a Meteor app to use as a test case that isn't also potentially bitrotted. I use the sample Python and PHP apps to test those stacks.

gischer:
I have several current Meteor apps I could use this on, as a test. And 
I’d be happy to do that.  However, that probably won’t happen for a few 
days, as I will be away from the keyboard for the weekend plus Monday.

------ Original Message ------

ocdtrekkie:
As you can see by our release schedule, we are in no rush. :) But we'd appreciate it!

ocdtrekkie:
meteor-spk is not close to useful for current Meteor apps, but I'm confident this PR is better than what was there, so I'm going to merge it in just to clear it off the board.

Issue #250, opened 2020-01-28
Plan to transition to contrib-buster64


ocdtrekkie:
We should look to upgrading vagrant-spk's default VM to buster: https://app.vagrantup.com/debian/boxes/contrib-buster64

As part of this, we should attempt to make stacks less dependent on specific versions of Debian, so that the scripts are likely to run successfully even if the user wishes to change out the box used and upgrades for new releases are less painful so that we can encourage the use of the `upgradevm` command intended for inclusion with 1.0.

ocdtrekkie:
If we cannot avoid references to something like stretch-backports or the like, we may want to consider detecting the box version as part of the stack setup script and injecting the correct version.

Examples of where this might come up currently:
https://github.com/sandstorm-io/vagrant-spk/blob/master/stacks/lemp/setup.sh#L10
https://github.com/sandstorm-io/vagrant-spk/blob/master/stacks/golang/setup.sh#L11

ocdtrekkie:
We know that the basics of vagrant-spk at least work fine with Buster, because @cdubz has already done it: https://github.com/babybuddy/babybuddy-sandstorm (Fun side note, `vagrant-spk upgradevm` after the command is merged would actually _downgrade_ this VM.)

cdubz:
For the record, I made this change because [Django 3.x only supports Python 3.6+](https://docs.djangoproject.com/en/3.0/releases/3.0/#python-compatibility) while [stretch has Python 3.5.x](https://packages.debian.org/stretch/python3). Moving to buster seemed more straightforward than trying a backport of some sort. So this issue is important for future Django support (though Django 2.x is fine on Python 3.5).

ocdtrekkie:
I would 100% agree on that being the more straightforward way to go. :) Fundamentally vagrant-spk should mostly work on any jessie and up Debian flavor, as long as the contrib- box with vboxsf support is used. We want to move in the direction of pushing more recent code in Sandstorm packages in the future, so we want to "unstick" a lot of it.

ocdtrekkie:
Fixed by #291

PR #291, opened 2021-03-19
Update base box to Debian Buster

ocdtrekkie:
I absolutely still need to test/validate this PR. :) But it addresses #250 in theory. I believe we can safely replace packages like php-fpm7.3 with php-fpm, as the latter depends on the former and we want to use the default version a given Debian release uses. Ideally, I need to figure out how to avoid so many version-specific references so switching boxes is not painful.

### DIFF
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -22,14 +22,14 @@ HOME=/etc/mysql /usr/sbin/mysqld --initialize
 
 # Spawn mysqld, php
 HOME=/etc/mysql /usr/sbin/mysqld --skip-grant-tables &
-/usr/sbin/php-fpm7.0 --nodaemonize --fpm-config /etc/php/7.0/fpm/php-fpm.conf &
+/usr/sbin/php-fpm7.3 --nodaemonize --fpm-config /etc/php/7.3/fpm/php-fpm.conf &
 # Wait until mysql and php have bound their sockets, indicating readiness
 while [ ! -e /var/run/mysqld/mysqld.sock ] ; do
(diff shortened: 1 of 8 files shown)


ocdtrekkie:
So, my dev machine is like, really hosed. :( Was totally going to test this tonight, but that is not gonna happen.

ocdtrekkie:
Hmmm... So I tried doing this on my Windows machine by just modifying the build scripts, got somewhere weird:

```
    default: Running: inline script
    default: /opt/app/.sandstorm/setup.sh: line 2: $'\r': command not found
    default: /opt/app/.sandstorm/setup.sh: line 5: $'\r': command not found
: invalid option name/.sandstorm/setup.sh: line 6: set: pipefail
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
Command failed with a non-zero exit status (1).
```

ocdtrekkie:
So that was a Windows/Unix line ending issue. ...I am not sure how that happened though. When I correct that in the files, this works for LEMP, which I've just tested.

ocdtrekkie:
@zenhack Even files I didn't write or edit are showing Windows line endings, I am wondering if GitHub Desktop did this to just my copy I downloaded, and if my PR actually is problematic or not.

With my Linux box out of commission, would you be willing to test this PR against the PHP app (lemp) and Python app (uwsgi) repos in the Sandstorm org?

zenhack:
@ocdtrekkie, will try to find time in the next day or two.

zenhack (approved):
LGTM. tested with both apps works as expected.

PR #292, opened 2021-03-27
Switch to Python 3

### LINKED ISSUE #247: vagrant-spk should be upgraded to Python 3

ocdtrekkie:
Fixes #247 

After making this change, on Ubuntu 20.04, I was able to complete a standard test of the php-app-to-package-for-sandstorm without issue. It probably would be good to test more of the various functions, but at minimum, this is basically useful without an unsupported version of Python, which we need to get away from.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -775,7 +775,7 @@ def shell(args):
         r'''for pid in $(pidof supervisor); do echo $pid $(cat /proc/$pid/status | grep -q 'Uid:.*'${SANDSTORM_UID} && echo ownership-correct || echo ownership-wrong) $(xargs -0 -n1 echo < /proc/$pid/cmdline  | grep -v -- - | head -n2 | tail -n1) $(grep -E -l ^PPid:[[:blank:]]*${pid}$ /proc/*/status | head -n1  | sed -r s,/proc/,,g | sed -r s,/status,,) ; done'''
         # get pid of supervisor, check process owner, get first child process
     )
-    splitted = data_from_guest.splitlines()
+    splitted = data_from_guest.decode('utf-8').splitlines()
     spk_pid = splitted[0]
     if spk_pid == 'no-spk':
         sys.stderr.write("No Sandstorm supervisor processes found.\n" +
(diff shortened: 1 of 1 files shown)


zenhack (approved):

ocdtrekkie:
So, enter-grain is indeed failing on Python 3, on a line sooner than I expected.

```
$ vagrant-spk enter-grain
ERROR: Failed to load enter_grain binary.Traceback (most recent call last):
  File "/usr/local/bin/vagrant-spk", line 1010, in <module>
    main()
  File "/usr/local/bin/vagrant-spk", line 1007, in main
    operation(args)
  File "/usr/local/bin/vagrant-spk", line 741, in shell
    desired_checksum = fd.read().split(' ')[0].strip()
TypeError: a bytes-like object is required, not 'str'
```

ocdtrekkie:
I got past the last error, hit raw_input, which was the expected issue, and now have a new one:

```
Looking for apps in dev mode...
Looking for grains...

Traceback (most recent call last):
  File "/usr/local/bin/vagrant-spk", line 1010, in <module>
    main()
  File "/usr/local/bin/vagrant-spk", line 1007, in main
    operation(args)
  File "/usr/local/bin/vagrant-spk", line 805, in shell
    grain_choice_human_readable = input(format_shell_grain_choices(supervisors)).decode('utf-8').strip()
  File "/usr/local/bin/vagrant-spk", line 261, in format_shell_grain_choices
    assert supervisors  # caller is supposed to bail out for us if this is empty
AssertionError
```

ocdtrekkie:
Okay, took a potshot guess at where I needed to decode for supervisors to have valid data, and I guess I was right! I am able to `vagrant-spk enter-grain` now on my Python3 vagrant-spk script. I confirmed this by catting number.txt in /var, which on the PHP sample app returns an integer of how many times the grain has been loaded.

zenhack:
Great, glad you figured it out.

PR #293, opened 2021-03-29
Release script pulls version from vagrant-spk

ocdtrekkie:
In review of #262, @zenhack suggested pulling this information from vagrant-spk rather than confirming it is the same in both places. This change to the script is tested (as a stub in a separate script file) to correctly do that. I flipped the order the variables are populated, because TAG_NAME is closer to the output of the vagrant-spk --version command. I attempted to figure out how to get TAG_NAME to populate in one line, but was unsuccessful, because I am a bash rookie. (Maybe I can argue it's more readable as two lines anyways...)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -34,11 +34,12 @@ function assert_git_state_is_clean() {
 }
 
 function get_release_name() {
-  # DISPLAY_VERSION gets used in the git tag description
-  DISPLAY_VERSION='1.0'
-
   # TAG_NAME gets used as the git tag name
-  TAG_NAME="v${DISPLAY_VERSION}"
+  TAG_NAME="$(./vagrant-spk --version)"
+  TAG_NAME="${TAG_NAME:12}"
+
+  # DISPLAY_VERSION gets used in the git tag description
+  DISPLAY_VERSION="${TAG_NAME:1}"
 }
 
 function assert_changelog_present() {
--- a/release.sh
+++ b/release.sh
@@ -35,8 +35,7 @@ function assert_git_state_is_clean() {
 
 function get_release_name() {
   # TAG_NAME gets used as the git tag name
-  TAG_NAME="$(./vagrant-spk --version)"
-  TAG_NAME="${TAG_NAME:12}"
+  TAG_NAME="$(./vagrant-spk --version | awk '{print $2}')"
 
   # DISPLAY_VERSION gets used in the git tag description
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -34,11 +34,12 @@ function assert_git_state_is_clean() {
 }
 
 function get_release_name() {
-  # DISPLAY_VERSION gets used in the git tag description
-  DISPLAY_VERSION='1.0'
-
   # TAG_NAME gets used as the git tag name
-  TAG_NAME="v${DISPLAY_VERSION}"
+  TAG_NAME="$(./vagrant-spk --version)"
```
zenhack, line 38: You can do `$(./vagrant-spk --version | awk '{print $2}')`, which I think is a bit more readable than the magical constant 12 below.

On master we still have the shebang for python 2, where it seems to print the version to stderr, so this won't work there since `$(...)` captures stdout. But it seems to do the right thing on python 3 and I don't think we care about the former -- but it's noteworthy since the python 3 change hasn't actually landed yet.

zenhack (commented):
One comment, otherwise LGTM

release.sh
ocdtrekkie: That is probably sensible. The advice I read suggested that what I did, not having used any outside programs, would technically be "faster". But awk is not slow, nor is the release script run often, and I would indeed prefer a single-line solution here.

Re: Python 2/3, my opinion is that we will not/should not push another release dependent on Python 2, and end users should not be running the release script, so yeah, I think I do not care? I was wondering if you felt maintaining Python 2 support was worth retaining, but I would hope that for app developers, being told they need Python 3 is not an exceptional request.

ocdtrekkie: Fixed by https://github.com/sandstorm-io/vagrant-spk/pull/293/commits/21a5a79c805be94a72eac5e3cb62eabe20ee2d3c

zenhack:
Quoting Jacob Weisz (2021-03-29 18:18:20)

>    Re: Python 2/3, my opinion is that we will not/should not push
>    another release dependent on Python 2, and end users should not be
>    running the release script, so yeah, I think I do not care? I was
>    wondering if you felt maintaining Python 2 support was worth
>    retaining, but I would hope that for app developers, being told
>    they need Python 3 is not an exceptional request.

I agree, let's consider Python 2 officially unsupported.

PR #308, opened 2021-09-13
Add missing .encode("UTF-8") calls.

zenhack:
Apparently I am the first person to run vagrant-spk upgradevm since the
Python 3 update. Without this, write() throws an exception complaining
that we passed it a str instead of bytes.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -664,7 +664,7 @@ def upgrade_vm(args):
     # Copy global setup script to e.g. install and configure sandstorm
     global_setup_script_path = os.path.join(sandstorm_dir, "global-setup.sh")
     with open(global_setup_script_path, "wb") as f:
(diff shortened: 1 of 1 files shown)


ocdtrekkie (approved):

ocdtrekkie:
Hmmm, you also adjusted a line in setupvm, but what about the one on line 672 in upgradevm that's similar?

I know I tested setupvm, but upgradevm is indeed less frequently used.

zenhack:
You're right, it's not necessary -- most likely I spotted it when skimming through, and assumed it needed the same fix, but it works just with the one change. I `--aemnded` it away.

I think it's needed for `global-setup.sh` because were opening it in binary mode? (`'wb'`)

ocdtrekkie (approved):

PR #314, opened 2021-11-29
Update bitrotten comment in Vagrantfile template.

zenhack:
The box itself was updated to buster a while back, but I noticed that
the comment still says stretch. I also got rid of the vboxfs bit; IIRC
there were separate images for these at one time, but they are no longer
needed.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -71,7 +71,7 @@ unless Vagrant::DEFAULT_SERVER_URL.frozen?
 end
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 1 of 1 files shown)


ocdtrekkie (approved):
It's worth noting that buster still requires the contrib- box, but since upgrading this should happen soon, and newer boxes a user might substitute will be without (such as bullseye), I see no reason not to remove the comment now.

ocdtrekkie:
Actually, a far older historical artifact is still in this comment: We no longer use our own snapshots of the official base box... we just use the official base box... Oh well.

PR #317, opened 2021-12-03
Minor formatting tweak in Vagrantfile message

zenhack:
This is a really trivial thing, but: my terminal's link detection seems
to want to include the period after the url in the url itself, which
means just trying to right click & open the link does not work
correctly. This patch gets around this by formatting the text in a way
that doesn't have the period at the end. I believe it still scans ok.


kpreid:
FYI, the traditional way to disambiguate this is to put angle brackets around the URL. Not saying that's better than the approach you took, just that angle brackets should be reliably a delimiter in any competent URL autolinker regardless of what other punctuation is present.

zenhack:
Makes sense; will keep that in my back pocket for later.

ocdtrekkie (approved):
I am not particularly set on ending punctuation on command line information anyways. It's not between sentences so I'm not concerned.

PR #326, opened 2022-05-09
Fix remote-fs.target dependency cycle

troyjfarrell:
Adding the automount option fixes a dependency cycle that can cause
Sandstorm to fail to start on boot for some development VMs.

The `host\x2ddot\x2dsandstorm.mount` is wanted by the `remote-fs.target`,
however it needs the `virtualbox-guest-utils.service` to mount
successfully.  The `virtualbox-guest-utils.service` can be made to run
before `remote-fs.target`, but I have never managed to make it be ready
for mounting Virtualbox Synced Folders at boot time.  It complains of
shared memory errors.  All of this causes the `remote-fs.target` to fail,
which causes `sandstorm.service` to fail to start at boot time.

This solution enables the automount option.  `/host-dot-sandstorm` is
instead no longer required for completion of the `remote-fs.target` and is
mounted on demand for `vagrant-spk`.  Everyone is happy.

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -142,7 +142,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     # /opt/app and /host-dot-sandstorm are used by vagrant-spk
     override.vm.synced_folder "..", "/opt/app"
-    override.vm.synced_folder ENV["HOME"] + "/.sandstorm", "/host-dot-sandstorm"
+    override.vm.synced_folder ENV["HOME"] + "/.sandstorm", "/host-dot-sandstorm", mount_options: ["x-systemd.automount"]
     # /vagrant is not used by vagrant-spk; we need this line so it gets disabled; if we removed the
     # line, vagrant would automatically insert a synced folder in /vagrant, which is not what we want.
     override.vm.synced_folder "..", "/vagrant", disabled: true


troyjfarrell:
Relevant documentation:

https://www.freedesktop.org/software/systemd/man/systemd.mount.html#x-systemd.automount

https://www.freedesktop.org/software/systemd/man/systemd.automount.html

ocdtrekkie:
Thanks. This sounds good though I feel out of my depth here. I'll test it as soon as I can. And also cc: @zenhack 

Do you know which sorts of flavors of development VMs are and aren't affected by this?

troyjfarrell:
This bug affects two of three `diy` VMs that I tested.  (I only have `diy` VMs.)  One of the two encounters the problem 50% of the time.  The second encounters the problem 90% of the time.

I don't understand why I need it for `/host-dot-sandstorm` but not for `/opt/app`.  I think the most likely explanation is a race condition, but I have no evidence for that.

I sent the PR because it solved a long-standing problem for me and I hoped that it would help others.  But if it's only happening on my system, you can close and ignore it.

ocdtrekkie:
Are you using a different base box by chance?

ocdtrekkie (approved):

ocdtrekkie:
Did a bit of testing with a project I am working on from a Windows machine, and this didn't throw any issues my way.

Issue #330, opened 2022-07-14
Strategy for upgrading MySQL 5.7 to 8.0?


ocdtrekkie:
So, I am looking at upgrading to bullseye64. I know it works for the lemp stack because I'm building an app with it, but the issue is upgrade-friendliness. Oracle offered 5.7 for Jessie, Stretch, and Buster, however, they are only offering MySQL 8.0 for Bullseye.

https://dev.mysql.com/blog-archive/inplace-upgrade-from-mysql-5-7-to-mysql-8-0/ says you can just start MySQL 8.0 on a 5.7 data directory, but I believe that is "assuming everything is healthy", which is not a good assumption for Sandstorm grains which we arbitrarily kill. (I am half-tempted to just PR updating everything, but I thought about it, and realized @zenhack would rightfully call me out on this one on review.)

The lemp and uwsgi stacks would both need a solution for this before we upgrade to Bullseye by default, and I am wondering if we can/should use the Nix solution used for TTRSS in general. I hate putting a TON of excess into stacks that may be used for greenfield projects, but maybe we could leave instructions to comment it out for new packages? (Or leave it commented out, and suggest people using `upgradevm` uncomment it?)

ocdtrekkie:
I guess this is also #276, but mysql5.5 to mysql5.7 was a long time ago, and mysql5.5 to mysql8.0 is now.

While thinking about this, I think the correct answer is to not use the upgrade code by default, since we want vagrant-spk to work out of the box for new packaging, and be efficient at doing that, but also we should probably include it, commented out, with a note to uncomment it if upgrading from an older version.

zenhack:
I think it's ok to upgrade this in the base boxes without adding upgrade logic, since upgradevm is a manual step anyway. Though we should maybe add a link to #276 in a comment as a warning. (We should separately come up with an upgrade strategy, but I don't think it needs to block the update for greenfield projects).

ocdtrekkie:
I am not sure where we would add a comment likely to be seen... Maybe when implementing something like #258?

I feel like your TTRSS upgrade logic would work for the actual upgrade strategy, but it probably needs to be relatively bundled up into: "uncomment these lines in setup and launcher if you previously released a version of this package using mysql5.5, and uncomment these ones if you previously released a package using mysql5.7".

But if you're comfortable not blocking on it, I will do some formal testing with the Python and PHP test apps and then mark the PR ready for review.

Issue #298, opened 2021-04-09
Create or designate test apps for Node, Meteor, and Go stacks


ocdtrekkie:
Generally, upon any radical vagrant-spk changes, I test against php-app-to-package-for-sandstorm and python-app-to-package-for-sandstorm, which are both excellent baseline test apps that use the base stack scripts for lemp and uwsgi respectively. (lesp is basically a simpler version of the lemp stack, so I also feel it is adequately tested this way.)

However, I do not have good test cases for our other stacks, which occasionally means they are broken until someone has a bad experience trying to use them. I think we need "Hello, world" level test apps that run with unmodified (or minimally modified) stack scripts for the Node, Meteor, and Go stacks.

zenhack:
Where do we want to collect these? I'd like to not just have half a dozen repos in random people's accounts that we have to separately keep track of, do we want to collect these hello world apps in a single repo somewhere?

...or I guess we could pull stuff in as submodules.

Ideally, I'd like to have this be semi-automated, but that's non-trivial given the combination of (1) interacting with grains programmatically is a bit of work and (2) launching VMs in CI would require a lot of setup and probably couldn't use GitHub actions (I wouldn't mind moving to something non-proprietary, but I do not want to shave this yak is the point).

ocdtrekkie:
In the short term, I'd maybe just throw together a list in a Markdown file in the repo. Since our current test apps are https://github.com/sandstorm-io/php-app-to-package-for-sandstorm/ and https://github.com/sandstorm-io/python-app-to-package-for-sandstorm/ they are pretty easy to find, and hence not listed anywhere.

I really just want something that's a reasonably good idea to validate any given stack against that's kinda "just the basics". Not sure if that would best be done by... literally just making more hello world apps like above or not. (But if so, we can transfer them to the sandstorm-io org)

ocdtrekkie:
Some of these are probably really good projects for newcomers to tackle, because it doesn't require deep understanding of Sandstorm, just how to build some small one-pager web apps in these platform stacks!

PR #331, opened 2022-07-14
Upgrade base box to Bullseye

### LINKED ISSUE #255: Update Node stack to 12, transition 10 to node10 stack
I am not in a rush to do this (marking it for vagrant-spk 1.1), but I figured I'd open the issue since people may have comments on it, particularly if they work with Node more than I do. In #242, we determined the "current LTS" would be `node` and that we'd retain our legacy Node stack, for apps that depend on it, as `node6`. Presumably then, now that Node's current LTS is 12, we should be planning to transition to it.

- Regarding Node 12: Sandstorm and Wekan both have/had some rough bits migrating to Node 12, I believe? How soon after an LTS becomes the LTS are we worried about having the stack ready in vagrant-spk?

- Regarding Node 10: Given above, is the breakage enough there's likely to be apps people want to package which still won't support Node 12? Given the delay in merging the new Node stack, there are probably no existing apps using it. Is there value in retaining it long term as `node10`?

- Regarding Node 6: As our "original" Node stack, most Node apps currently in Sandstorm were probably built with this. But how long should we retain it as a vagrant-spk stack? Are there a lot of apps which still cannot work with newer Node versions that we either want to update or package in the future?

ocdtrekkie:
I think we need to address #330 before we merge this, and I also am going to do some testing against our test packages yet. We still need good test packages for some stacks (#298), which is also arguably blocking #290.

This also fixes #255. Node 14 is a supported LTS releae, and node-capnp supports Node 14, so we'll bump that up to be relatively modern as well.

### DIFF
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -33,10 +33,10 @@ HOME=/etc/mysql /usr/sbin/mysqld --initialize
 
 # Spawn mysqld, php
 HOME=/etc/mysql /usr/sbin/mysqld --skip-grant-tables &
-/usr/sbin/php-fpm7.3 --nodaemonize --fpm-config /etc/php/7.3/fpm/php-fpm.conf &
+/usr/sbin/php-fpm7.4 --nodaemonize --fpm-config /etc/php/7.4/fpm/php-fpm.conf &
 # Wait until mysql and php have bound their sockets, indicating readiness
 wait_for mysql /var/run/mysqld/mysqld.sock
-wait_for php-fpm7.3 /var/run/php/php7.3-fpm.sock
+wait_for php-fpm7.4 /var/run/php/php7.4-fpm.sock
 
 # Start nginx.
 /usr/sbin/nginx -c /opt/app/.sandstorm/service-config/nginx.conf -g "daemon off;"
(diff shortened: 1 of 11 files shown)


ocdtrekkie:
PHP test app is good, I submitted fixes for other issues unrelated to the base box upgrade.

Python stack is currently broken, with `default: E: Package 'python-virtualenv' has no installation candidate`. I think it's probably a safe assumption that's a Python 2 vs. 3 issue, which means we probably need to update our sample app for Python 3. I'm not looking at that tonight, but I'll document it here as a blocker to merging this PR.

ocdtrekkie:
Okay, @zenhack, this is ready for your review. Note that on my Windows box, I had to run `virtualenv --copies` in build.sh to avoid the symlink issue, but I don't feel we should change that by default because it can be very wasteful in packaging output.

https://github.com/sandstorm-io/python-app-to-package-for-sandstorm/pull/4 is the simultaneously-needed update to our test app.

ocdtrekkie:
Hmmm, I just pushed that last commit on here for Node, before realizing we waited last time until node-capnp was supported. I don't think apps actually necessarily *need* node-capnp, and we don't currently have a test app either for that stack, so I think for the moment, I would be in favor of just keeping current as Node 14 goes EOL in a handful of months, and we push actual vagrant-spk releases rarely.

zenhack (approved):
LGTM.

I agree with keeping node current, rather than waiting for node-capnp to catch up.

PR #351, opened 2023-11-05
Change base box to Bookworm

### LINKED ISSUE #343: Upgrade to Debian Bookworm
We need upgrade our default machine and our stacks to Debian Bookworm.

This entails ensuring especially that our stacks are updated as necessary for the correct versions of everything that refers to the distribution, and we need to ensure the PHP and Python test apps still work correctly.

https://github.com/sandstorm-io/php-app-to-package-for-sandstorm
https://github.com/sandstorm-io/python-app-to-package-for-sandstorm

Our last upgrade, to Bullseye, was here: #331

ocdtrekkie:
Fixes #343 

This is not tested/validated yet, but "should work".

### DIFF
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -41,10 +41,10 @@ HOME=/etc/mysql /usr/sbin/mysqld --initialize
 
 # Spawn mysqld, php
 HOME=/etc/mysql /usr/sbin/mysqld --skip-grant-tables &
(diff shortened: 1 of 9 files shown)


ocdtrekkie:
This PR is now tested on the PHP and Python test apps.

@troyjfarrell Does the fix I made for uwsgi seem correct to you?

troyjfarrell:
> @troyjfarrell Does the fix I made for uwsgi seem correct to you?

I did not test it, but it looks correct.

PR #352, opened 2024-03-09
Initalize MySQL Once

ocdtrekkie:
So I believe this change is necessary to stabilize our lemp and uwsgi stacks, and fixes the roadblock I've been fighting on my own package. MySQL's documentation states --initialize should exit without changes if the directory isn't empty (aka, already initialized), but my guess is because we launch initialize without a password (or skip-grant-tables like our main process launch), it encounters an access error trying to verify the initialization is already done. I think the best choice is to just only initialize a grain once, it avoids some log cruft and is (probably) more efficient.

https://dev.mysql.com/doc/refman/8.0/en/data-directory-initialization.html#data-directory-initialization-procedure

I am also taking a behavior Ian used effectively in his grains and introducing it as a standard behavior, both for initializing and populating the database. Since the latter isn't always required, I'm including it in the template commented out.

### DIFF
--- a/stacks/lemp/launcher.sh
+++ b/stacks/lemp/launcher.sh
@@ -35,15 +35,26 @@ for f in $log_files; do
     fi
 done
 
-# Ensure mysql tables created
-# HOME=/etc/mysql /usr/bin/mysql_install_db
-HOME=/etc/mysql /usr/sbin/mysqld --initialize
+if [ ! -e /var/.db-initialized ]; then
+    # Ensure mysql tables created
+    HOME=/etc/mysql /usr/sbin/mysqld --initialize
+    touch /var/.db-initialized
+fi
 
 # Spawn mysqld, php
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
@orblivion @troyjfarrell If either of you were able to verify that a grain built with the php or uwsgi test apps can only launch once (aka, fail, if you let the grain shut down and attempt to reopen it) off master, but work correctly if you create them using this PR, I would really love to know I am not crazy.

ocdtrekkie:
So I both simplified this and ensured it should work now even if the original grain wasn't created with this in place. Instead of initializing and leaving a file to check... it simply checks for a folder that should only exist if mysql has been initialized already.

orblivion:
I tried both the php-app-to-package-for-sandstorm and python-app-to-package-for-sandstorm. With master they would initially load, but fail to load if I restarted the grain. With this branch they would still load after restarting the grain.

orblivion:
(we're doing QA here so I'm being specific and redundant)

PR #353, opened 2024-03-16
Rename publisher in Windows installer

ocdtrekkie:
I'm not including this in 1.2, but leaving this PR out there in part as a reminder to do it next release. There is no Sandstorm Development Group, Inc. so we should rename this. It shows up in the legacy Control Panel primarily.

### DIFF
--- a/windows-support/windows-installer.iss
+++ b/windows-support/windows-installer.iss
@@ -3,7 +3,7 @@
 
 #define MyAppName "vagrant-spk"
 #include "state/version.iss"
-#define MyAppPublisher "Sandstorm Development Group, Inc."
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
One thing that bugs me is the installer refers to an "app URL" for support and updates, and it points at docs.sandstorm.io. I want to change this, but the docs is probably the best place for support, and the GitHub would be the best place for updates, but I don't want to put a GitHub link there or have to add a separate URL to the config for this.

I am sort of wondering if we should have a "landing page" for some of our project tools on sandstorm.org with some of the basic information and links. Something like a sandstorm.org/vagrant-spk which links to the latest downloads, the right docs files, etc.

PR #355, opened 2025-02-08
Go: Get latest version correctly

ocdtrekkie:
So our Go stack was broken, and I figured out why: https://go.dev/VERSION?m=text used to return the latest Go version (technically, the version the Go website uses, but I digress). Now it also includes a timestamp, and that broke this script.

### DIFF
--- a/stacks/golang/setup.sh
+++ b/stacks/golang/setup.sh
@@ -12,7 +12,7 @@ if [ -e /opt/app/.sandstorm/go-version ]; then
     curl -L "https://go.dev/dl/$(cat '/opt/app/.sandstorm/go-version').linux-amd64.tar.gz" -o go.tar.gz
 else
     # Get the newest version for a new project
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
@orblivion Can I get your review?

orblivion:
It seems fine assuming it works.

orblivion:
In the long run: I might suggest something a little less error prone but this change doesn't make it worse. Also I'm a fan of sig checking but apparently that's not available for Go. I usually put the hash in my scripts instead but that targets a specific version. More recently, Go added a thing called "toolchains" where it can start with 1.21 and update itself from there like a dependency, but that requires building which will be slow.

Maybe whatever Troy comes up with for Tempest would be useful.

ocdtrekkie:
There is an alternative way to query a JSON of Go versions but I figured this was simpler. We are downloading Go from the same server, so if curl fails here, the Go download will likely fail anyways as well.

Issue #356, opened 2025-05-15
The 'vagrant-spk init' command should warn the user before clobbering changes made to an existing package


neuroradiology:
If a user follows the workflow from the readme

```
git clone git://github.com/sandstorm-io/vagrant-spk
git clone git://github.com/sandstorm-io/php-app-to-package-for-sandstorm
export PATH=$(pwd)/vagrant-spk:$PATH
cd php-app-to-package-for-sandstorm
vagrant-spk setupvm lemp
vagrant-spk vm up
vagrant-spk init
# edit .sandstorm/sandstorm-pkgdef.capnp in your editor of choice
vagrant-spk dev
# visit http://local.sandstorm.io:6090 in a web browser
# log in as Alice, the admin account
# launch an instance of the example app, play around with it
# then, press Ctrl-C to stop the tracing vagrant-spk dev

```

and runs `vagrant-spk init` again at any point, it will clobber the changes they have made (to the sandstorm-pkgdef.capnp file, in particular). It would be courteous to warn the user about this before executing the command; ideally with a check to see if such a warning is appropriate.

ocdtrekkie:
This is very related to #258 but I agree this should be covered as well.

PR #360, opened 2025-10-30
Require confirmation for setupvm, upgradevm, init

### LINKED ISSUE #258: setupvm and upgradevm should require confirmation
This is particularly egregious with setupvm, but I'd like to fix it for both. These commands overwrite existing app packaging files without any confirmation. In the case of setupvm it is basically a reset of your packaging efforts, upgradevm it is just going to force you to fix your app build for the latest VM.

Either way, we should probably put a quick "are you sure?" in here. We may also need to look at a flag or environment variable to skip these confirmations. While nobody is using it AFAIK, `vagrant-spk auto` is a thing that exists in the code that @paulproteus was building, and other situations where people have scripted calling setupvm may exist. (Nobody is currently scripting calling upgradevm yet because it doesn't exist yet.)

ocdtrekkie:
Fixes #258 and #356 

So this was initially written by Perplexity AI (because I thought this would be a good test case for upgrading existing code in an easily-parsable and testable way), but I stripped some changes back (it messed with a variable name), and it handled init differently which seemed silly, so I made it handle init the same way it handles setupvm and upgradevm. It also included the exact file being checked in the prompt, and I changed the language there because I am testing for the existence of one file, but using it to prompt for confirmation of the entire action. (I believe the `return` should end the entire function?)

Note that I have _not yet tested this at all_, but will do some tests on it later today and verify if it works as intended. It reads reasonable to me though.

I'd be thrilled if anyone else was able to test this too. cc: @neuroradiology @orblivion @troyjfarrell

### DIFF
--- a/vagrant-spk
+++ b/vagrant-spk
@@ -395,6 +395,14 @@ def ensure_host_sandstorm_folder_exists():
         with open(keyring_file, "wb") as f:
             pass
 
+def confirm_overwrite(filepath, noninteractive):
+    if noninteractive:
+        return True
+    if os.path.exists(filepath):
+        resp = input("This command will overwrite some existing packaging configuration. Proceed? [y/N]: ").strip().lower()
+        return resp == 'y'
+    return True
+
 class StackPlugin(object):
     def __init__(self, plugin_name):
         self._plugin_name = plugin_name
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I am going to merge and if we want to adjust the criteria there I will do so separately.

